PR #20, opened 2017-07-20
Merge to master

ketzacoatl:
I'd rather ff-merge this instead of having the merge commit, but I really want to get some review on these changes before I do that. cc @lehins and @borsboom.

The commit list is pretty clear about the changes, so rather than repeat them entirely, I'll highlight a few key changes:
* the `vpc` module has been renamed to `vpc-legacy`
* `subnets` has been refactored to create only one type of subnet at a time, not a mix of public and private subnets (this works around an issue where changing from 0 public subnets to something will error out in terrible ways)
* various variable updates to the `asg` module, though I think I'm the only one using it


ketzacoatl:
Manny and Alexey have reviewed and signed off on this (via slack), so I've merged the changes to master and this PR is good to close.

Issue #17, opened 2017-03-31
add a tutorial to demonstrate how to use credstash


ketzacoatl:
The `credstash` module is a great addition, but it needs some docs to be really useful.

PR #21, opened 2017-08-07
Credstash doc and improvements

lehins:
This PR contains a rewrite of credstash modules and a detail documentation on how to use them.
Modules that depend on credstash (logstash, elasticsearch, etc) have also been adjusted.


docs/credstash.md
```diff
@@ -0,0 +1,215 @@
...
+assures that any EC2 instance assuming this role can use credstash only for
+reading and/or writing secrets, but cannot do anything else with neither KMS
+Master Key nor DynamoDB table.
+
+Here is how to give read only permission to that role:
+
+```
+module "credstash-grant" {
+  source            = "github.com/fpco/fpco-terraform-aws/tf-modules/credstash-grant"
+  kms_key_arn       = "${data.terraform_remote_state.credstash.kms_key_arn}"
```
ketzacoatl: If you recommend the `credstash-setup` module _always_ be in another env, we should state that. 

If putting the `credstash-setup` module in another env is often a good idea or helpful, we might want to mention that.

Either way, the doc ought to make note that this example assumes there is a data block for the remote state where the `credstash-setup` module is (or include a short snippet in the example).

ketzacoatl: oops, I didn't mean to "start a review", I just wanted to add a comment..

docs/credstash.md
```diff
@@ -0,0 +1,215 @@
...
+  reader_policy_arn = "${data.terraform_remote_state.credstash.reader_policy_arn}"
+  roles_count       = 1
```
ketzacoatl: what would it look like if `roles_count` was > 1?

tf-modules/credstash-grant/main.tf
```diff
@@ -0,0 +1,83 @@
...
+    command = "${path.module}/grant.sh create writer ${var.writer_context == "" ? "" : "--context ${join(",", split(" ", var.writer_context))}"} ${var.kms_key_arn} ${var.roles_arns[count.index]}"
+  }
+
+  provisioner "local-exec" {
+    when       = "destroy"
+    on_failure = "continue"
+    command    = "${path.module}/grant.sh revoke ${var.kms_key_arn} ${var.roles_arns[count.index]}"
+  }
+}
+
```
ketzacoatl: I love seeing the details you worked out here, this is great!

tf-modules/credstash-setup/main.tf
```diff
@@ -41,41 +76,61 @@ resource "aws_dynamodb_table" "credstash-db" {
...
+  ]
 }
-
-
-//KMS Key ARN. It can later be used to store and retrieve secrets:
-// credstash put -k kms_key_arn secret_key secret_value
-// credstash get -k kms_key_arn secret_key
-output "kms_key_arn" {
-  value = "${aws_kms_key.credstash-key.arn}"
+END_POLICY
```
ketzacoatl: I believe this is a place where I'd recommend using the https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html instead of the JSON policy. If it's possible, but not required.

tf-modules/credstash-setup/variables.tf
```diff
@@ -27,3 +31,14 @@ variable "db_table_name" {
...
+variable "create_reader_policy" {
+  default = false
+  description = "Should credstash Secret Reader IAM Policy be created."
+}
+
+variable "create_writer_policy" {
+  default = false
+  description = "Should credstash Secret Writer IAM Policy be created."
+}
+
```
ketzacoatl: Q: does it make any sense for a user to create their own reader/write policy?

Q: Would any of this code need to change if we needed to give an EC2 instance a credstash role, _AND_ another role to do something else on AWS (say, access an EBS volume, S3 bucket, etc) ?

tf-modules/elasticsearch/iam.tf
```diff
@@ -22,37 +22,18 @@ resource "aws_iam_role_policy_attachment" "data-node-attach-ec2-discovery" {
...
-  count = "${var.data_node_count}"
-  role = "${element(aws_iam_role.data-node-role.*.name, count.index)}"
-  policy_arn = "${module.credstash-reader.reader_policy_arn}"
+module "credstash-grant" {
+  source            = "../credstash-grant"
+  kms_key_arn       = "${var.credstash_kms_key_arn}"
+  reader_policy_arn = "${var.credstash_reader_policy_arn}"
+  roles_count       = "${var.master_node_count + var.data_node_count}"
+  roles_arns        = ["${concat(aws_iam_role.master-node-role.*.arn, aws_iam_role.data-node-role.*.arn)}"]
+  roles_names       = ["${concat(aws_iam_role.master-node-role.*.name, aws_iam_role.data-node-role.*.name)}"]
```
ketzacoatl: ah, I guess this might be a good example to mention / use in the docs (for multiple roles).

ketzacoatl:
I added one comment about the remote state data, but otherwise that doc, and the related module updates, both look great!

ketzacoatl:
One other update I would request: can you include an example that shows how to use credstash to store/retrieve config files as you do with logstash?

tf-modules/credstash-setup/variables.tf
lehins: A: Those are minimal policies to be able to read/write from DynamoDB table. Use can create less restrictive policies and attach to the same role created by the user if needed, all information necessary for that is exported by the module.
A: Many policies can be attached to a single role, that's why it is not a role that is created in this module, but the policies, which can be used later if needed. So, no, nothing here would need to be changed if a user wants to access EBS volume S3, what have you, he can create another policy and attach to the same role.

tf-modules/credstash-grant/main.tf
lehins: Nice, thank you.

docs/credstash.md
lehins: `roles_count` has to match the length of `roles_arns` and `roles_names`, if it's higher it will result in an error, if less, well, not all roles will be used. I really wish this wan't necessary, but unfortunately if roles are created at the same time this module is used, terraform will not be able to evaluate length of roles and use it as a count. I probably should add a link to the terraform issue about this.

docs/credstash.md
lehins: `credstash-setup` is meant to be a totally separate environment (hence usage of `terraform_remote_state`). You are totally right, I'll give more detail on that.

tf-modules/elasticsearch/iam.tf
lehins: Will do

lehins:
Thank you for the comments, I appreciate it. I'll address the comments and add certstrap example as well.

ketzacoatl:
Once this PR is complete, I believe it will fulfill the goal in #17.

PR #23, opened 2017-09-12
Extracted Route53 attachments into separate module for ELK componenets

lehins:
ELK will assume that DNS records created in some other way if Route53 isn't used


ketzacoatl:
That looks great! :+1: Merged

PR #24, opened 2017-09-15
Adds packer for kubespray base AMI building

denibertovic:


denibertovic:
Awesome feedback @ketzacoatl :+1:  
I've addressed all of the above so I'm merging.

PR #29, opened 2017-10-06
A single-node autoscaling group running Gitlab

NorfairKing:


ketzacoatl:
In general, we would put new modules in `tf-modules` within this repo, and then a _complete_ TF environment (_testing_ the modules) would go in the `tests` path. Given the details with Gitlab here, it's sufficiently simple enough that a module for the ASG + init may or may not be worth it. My thought was to instead have a test env that uses the single-node ASG module the way you have with the init needed to run gitlab in that test.

In otherwords, if adding a test env here, I would have expected the code from https://gist.github.com/NorfairKing/f7bd0ce362fbe1d9b132551d8167b3ae plus the code in the "module" from this PR, rather than a module defined in the tests and some outside env that uses the module.

It's confusing, so I'm not sure how much sense my description makes.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+  name                    = "test"
+  az                      = "${var.az}"
+  key_name                = "${var.key_name}"
+  key_file                = "${var.key_file}"
+  ami                     = "${var.instance_ami}"
+  instance_type           = "${var.instance_type}"
+  name_suffix             = "gitlab-snasg"
+  subnet_id               = "${var.subnet_id}"
+  security_group_ids      = ["${aws_security_group.snasg-security-group.id}"]
+  region                  = "${var.region}"
```
ketzacoatl: you can also use the `aws_region` `data` source to retrieve details like this, see https://www.terraform.io/docs/providers/aws/d/region.html

```
data "aws_region" "current" {
  current = true
}
```

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+mount /dev/xvdf1 /gitlab
+
+cp /etc/fstab /etc/fstab.orig
+echo "LABEL=gitlab            /gitlab  ext4   defaults,nofail     0 2" >> /etc/fstab
+
+apt-get install -y docker docker.io
+cmd="docker run --detach \
+    --publish 443:443 \
+    --publish 80:80 \
+    --publish 8022:22 \
```
ketzacoatl: might want to make these ports configurable (as variables), or at least the SSH port.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+    --restart always \
+    --volume /gitlab/config:/etc/gitlab \
+    --volume /gitlab/logs:/var/log/gitlab \
+    --volume /gitlab/data:/var/opt/gitlab \
+    gitlab/gitlab-ce"
+echo "$cmd" > /etc/rc.local
```
ketzacoatl: I like this trick, sweet!  `docker create` and `docker start` are worth mention, though I don't remember if that survives a reboot.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+ * parted --script /dev/xvdf -- mklabel msdos
+ * parted --script /dev/xvdf -- mkpart primary 0 -1
+ * mkfs -t ext4 -F /dev/xvdf1
+ * e2label /dev/xvdf1 gitlab
+ *
+ * After running the above code to initialise the EBS, terminate the instance
+ * and the autoscaling group will bring up a new instance that will be running
+ * gitlab once it is done initialising.
+ */
+
```
ketzacoatl: you might want to publish some outputs, though I haven't thought much about which you ought to publish.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+    --volume /gitlab/config:/etc/gitlab \
+    --volume /gitlab/logs:/var/log/gitlab \
+    --volume /gitlab/data:/var/opt/gitlab \
+    gitlab/gitlab-ce"
+echo "$cmd" > /etc/rc.local
+$cmd
+END_INIT
+}
+
+resource "aws_security_group" "snasg-security-group" {
```
ketzacoatl: you can generally omit the `-security-group` from naming this resource. `gitlab-asg` or `gitlab` would be fine, as it would be referred to as `${aws_security_group.gitlab-asg.foo}`, so including the "type" ("security-group") in that is redundant / not helpful.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+  name        = "snasg-security-group"
```
ketzacoatl: note that this `name` is what you would end up seeing in the AWS console, so this ought to be something that helps the user pick this needle out of the haystack (you can easily end up with a _lot_ of SG in an account).

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+ * parted --script /dev/xvdf -- mkpart primary 0 -1
+ * mkfs -t ext4 -F /dev/xvdf1
+ * e2label /dev/xvdf1 gitlab
+ *
+ * After running the above code to initialise the EBS, terminate the instance
+ * and the autoscaling group will bring up a new instance that will be running
+ * gitlab once it is done initialising.
+ */
+
+module "snasg" {
```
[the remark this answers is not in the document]
NorfairKing: fixud.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+  source                  = "../../tf-modules/single-node-asg"
+  name                    = "test"
+  az                      = "${var.az}"
+  key_name                = "${var.key_name}"
+  key_file                = "${var.key_file}"
+  ami                     = "${var.instance_ami}"
+  instance_type           = "${var.instance_type}"
+  name_suffix             = "gitlab-snasg"
```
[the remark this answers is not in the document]
NorfairKing: fixed.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+ * e2label /dev/xvdf1 gitlab
+ *
+ * After running the above code to initialise the EBS, terminate the instance
+ * and the autoscaling group will bring up a new instance that will be running
+ * gitlab once it is done initialising.
+ */
+
+module "snasg" {
+  source                  = "../../tf-modules/single-node-asg"
+  name                    = "test"
```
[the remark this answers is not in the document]
NorfairKing: fixed.

tests/gitlab-single-node-asg/main.tf
```diff
@@ -0,0 +1,116 @@
...
+  az                      = "${var.az}"
+  key_name                = "${var.key_name}"
+  key_file                = "${var.key_file}"
+  ami                     = "${var.instance_ami}"
+  instance_type           = "${var.instance_type}"
+  name_suffix             = "gitlab-snasg"
+  subnet_id               = "${var.subnet_id}"
+  security_group_ids      = ["${aws_security_group.snasg-security-group.id}"]
```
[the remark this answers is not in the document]
NorfairKing: fixed.

tests/gitlab-single-node-asg/main.tf
NorfairKing: I'm not sure whether this is a good idea, given that you then 1. cannot have things in multiple regions at the same time and 2. include an assumption that the availability zone you supply is in this region.

Let me know if you still recommend this `data` approach, and I'll fix it.

tests/gitlab-single-node-asg/main.tf
NorfairKing: Fixed. I tried also making it configurable _whether_ they are published, but it doesn't seem possible to put a colon (`:`) in the values of the ternary operator in terraform code.

tests/gitlab-single-node-asg/main.tf
NorfairKing: fixed.

tests/gitlab-single-node-asg/main.tf
NorfairKing: fixed

ketzacoatl:
Looks great to squash/rebase/ff-merge!

NorfairKing:
I rebased and squashed manually.

PR #40, opened 2017-11-25
Enhanced gitlab-asg module, added output to the nat module

synesisinc:
* Added output for the nat ec2 instance private ip address
* Added `load_balancers` and `public_ip` parameters to the gitlab-asg  module


ketzacoatl:
These updates have been merged, this is good to close.

PR #41, opened 2017-11-28
Cloudwatch auto recover existing e2 instance

kerscher:
@ketzacoatl, this updates the EC2 instance to use the new module that was split out from it. You should probably take a look at both. Parameters for the new module were devised based on the one here, so it should be unchanged for users.


ketzacoatl:
When testing this module in practical use, I see the need for using `-target` to get across the cross-module count calculation issue with TF modules. I think it makes sense to hold off on this PR until TF has worked out the count/interpolation/calculation issue across modules.

PR #42, opened 2017-11-29
Cloudtrail logs

kerscher:
Adds a `cloudtrail` module that creates a S3 bucket and sends CloudTrail logs towards it for the current region. Since `aws_cloudtrail` resources on Terraform use the currently active region, S3 buckets default to using callee’s. This module includes policies for writing out to the bucket as `data.aws_iam_policy_document` to ensure (as much as possible) compile-time verification.

Optionally, if user provides `kms_key_arn` contents are encrypted.

### DIFF
--- a/tf-modules/cloudtrail/variables.tf
+++ b/tf-modules/cloudtrail/variables.tf
@@ -1,0 +1,22 @@
+variable "name_prefix" {
+  description = "name to be prepended to all resources"
+}
+
+variable "aws_cloud" {
+  description = "set to 'aws-us-gov' if using GovCloud, otherwise leave the default"
+  default     = "aws"
+}
+
+variable "include_global_service_events" {
+  description = "boolean, if true logs IAM events as well as region-specific ones"
+  default     = true
+}
+
+variable "enable_logging" {
+  description = "boolean, defines if logging should be started or stopped"
+  default     = true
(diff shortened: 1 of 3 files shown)


tf-modules/cloudtrail/variables.tf
```diff
@@ -0,0 +1,27 @@
...
+  default     = ""
+}
+
+variable "include_global_service_events" {
+  description = "boolean, if true logs IAM events as well as region-specific ones"
+  default     = true
+}
+
+variable "enable_logging" {
+  description = "boolean, defines if logging should be started or stopped"
```
ketzacoatl: I'm not sure what "logging" this is referring to, maybe either explain that, or go more towards: "Boolean, maps to `aws_cloudtrail`'s `enable_logging` parameter"

ketzacoatl:
I left one comment, not a big issue. The module looks good to me, simple and clear.

tf-modules/cloudtrail/variables.tf
kerscher: Agreed. I did spend a while looking at the screen when I wrote this one. Your solution is straightforward, and I added it in the commit below 👇

ketzacoatl:
:+1:

PR #44, opened 2017-12-05
update s3-remote-state documentation

jn9999:
Tried this module but discovered the documentation to not match how the module works now.

> Initializing modules...
> - module.s3-remote-state-bucket
> Error getting plugins: module root: 
>     module s3-remote-state-bucket: iam_users is not a valid parameter
>     module s3-remote-state-bucket: iam_groups is not a valid parameter
>

Not sure what an example using principals instead should look like.


ketzacoatl:
Great catch, and thanks for the contribution!

PR #45, opened 2017-12-07
cloudwatch-logs-s3-bucket

kerscher:
Adds a module with documentation for creating S3 buckets for CloudWatch Logs, with necessary IAM.

User still has to create the Log Groups elsewhere and configure instances, as each caller might have different use cases and types of streams in mind.


kerscher:
Can we merge this?

Issue #66, opened 2018-03-06
add IAM modules to kube-stack test env


ketzacoatl:
Testing the `kube-stack` module should include the IAM profiles needed. We already have the modules, so adding this looks like:

```
# IAM profiles for kube controllers/workers
module "kube-controller-iam" {
  source      = "../modules/kube-controller-iam"
  name_prefix = "${var.name_prefix}"
}

module "kube-worker-iam" {
  source      = "../modules/kube-worker-iam"
  name_prefix = "${var.name_prefix}"
}
```

..and then passing to the `kube-stack` module:

```
  ...
  controller_iam_profile   = "${module.kube-controller-iam.aws_iam_instance_profile_name}"
  ...
  worker_iam_profile       = "${module.kube-worker-iam.aws_iam_instance_profile_name}"
  ...
```

Issue #67, opened 2018-03-06
update handling of security groups in kube-stack test env


ketzacoatl:
The security group modules now create a base SG (for each node/role), and a bunch of rules are then added to that base SG. The `kube-stack` test env - https://github.com/fpco/terraform-aws-foundation/tree/master/examples/kube-stack-private for example - has not been updated since those SG modules were refactored. After this update, the test env ought to deploy cleanly and work in a basic sense.

ketzacoatl:
I've updated the `kube-stack-private` env, and that is complete. The public env still needs the same updates. I'm reassigning this to @lpaulmp to complete those updates as part of his work with this kube-stack.

PR #68, opened 2018-03-06
`kube-stack-*` security groups and IAM modules

kerscher:
Closes issues #66 and #67


ketzacoatl:
With the sec-groups now using rule resources, there are a few ways to do this. Here is a slight alternative:
* define a single security group for each role (in this case worker/controller/lb) - there's a the "security-group-base" module for this
* use rule modules to attach SG rules to the base SGs, including the SSH/egress/etc

In practice, it looks like this for a role: https://github.com/ketzacoatl/compute-demo/blob/e39b91fa4b7ae595cc8a3dc921dfdfa11693930f/terraform/vpc/workers.tf#L1-L36

I'd also recommend keeping the Security Groups with each roles, rather than move them to `main.tf`.

kerscher:
I moved resources around so it looks more like how it’s organized in actual environments. Security groups live in `name-of-module/firewall.tf`.

As for the security groups refactor, I did not move them as I don’t think the changes are beneficial:

- Security group rule modules don’t take lists → every SSH rule module has to be duplicated for each specific security group.
- `security-group-base` only wraps `aws_security_group` without any other changes → changing is a no-op

ketzacoatl:
Rather than fumble in explaining what I have in mind for the test/demo env, I have made the updates directly. The `kube-stack-private` env has been updated and initial tests were successful, I will leave @lpaulmp to run additional testing as part of his own work on these env. I am also leaving the `kube-stack-public` updates to @lpaulmp to make. This PR is good to close, Paul will make the remaining update to the public env in a separate PR. Thanks to all for making this happen!

PR #72, opened 2018-03-09
Adds cloudwatch exporter iam

denibertovic:
Same as with dnscontroller-iam. 

Also, adds a tiny fix for that role (dnscontroller) as well, to make it more consistent. 

Specifically the naming of the outputs should not be "some_elaborate_name_FOO" it should just be "foo" since the user will access it as "module.mydnscontroller.FOO". That's my thinking anyway. The only thing I'm not sure off is of the outputs should be `role_name` and `role_arn` or is `name` and `arn` good enough. 
What do we usually do in these cases?

### DIFF
--- a/modules/cloudwatch-exporter-iam/README.md
+++ b/modules/cloudwatch-exporter-iam/README.md
@@ -1,0 +1,20 @@
+# Prometheus Cloudwatch exporter IAM role
+
+This module creates an `aws_iam_role` that get's assigned to the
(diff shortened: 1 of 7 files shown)


ketzacoatl:
`role_arn` is usually preferable to `arn`, especially for IAM resources.

denibertovic:
@ketzacoatl  added `role_` to outputs. Is this good to merge?

PR #73, opened 2018-03-11
Brought back ability to use ELBs for Elasticsearch and Kibana in ELK

lehins:
* Using ALBs is still possible, but ELBs will be created as an option.
* Extracted away ACM certificates lookup, those are now expected as a variable
  to external LBs.
* Elasticsearch is now properly deployed in private subnets, while external
  LB is in public

Here is an example difference of how to use ELB vs ALB for Elasticsearch (variable to `elk-stack` module):
```hcl
  elasticsearch_internal_alb              = {
    "arn"               = "${aws_alb.es-internal.arn}"
    "dns_name"          = "${aws_alb.es-internal.dns_name}"
    "zone_id"           = "${aws_alb.es-internal.zone_id}"
    "security_group_id" = "${aws_security_group.es-internal-alb.id}"
  }
```
So, in order to deploy ELB, all that is necessary is `deploy_elb`:
```hcl
  elasticsearch_internal_alb              = {
    "deploy_elb"        = true
    "security_group_id" = "${aws_security_group.es-internal-alb.id}"
  }
```

Few extra arguments are also available in that map:
  * `security_group_id` (required)
  * `certificate` (required) - ARN of ACM certificate to be used.
  * `deploy_elb` (optional, default: false) - if set to true ELB will be deployed.
  * `deploy_elb_internal` (optional, default: true) - if set to false ELB will be deployed as an external one.
  * `deploy_elb_cross_zone` (optional, default: true) - if set to false cross AZ for ELB will be off.

Below are only required if `deploy_elb` is set to `false` or omitted:
  * `arn` - ALB ARN
  * `dns_name` - host-header for ALB listener rule

Same approach goes for all three places where ALB/ELB is used:
* Elasticsearch internal API (for unauthenticated requests from logstash running in the same VPC)
* Elasticsearch external API with BasicAuth
* Publicly accessible Kibana with BasicAuth.

### DIFF
--- a/modules/elasticsearch/data-nodes.tf
+++ b/modules/elasticsearch/data-nodes.tf
@@ -3,9 +3,10 @@
  *
  * Just as with master nodes, each data node has a persistent EBS volume created for.
  * Data nodes also function as coordinators, thus each one has 9200 port open for the API.
- * Elastic load balancer is created for elasticsearch API enpoint. Every node is managed
- * by an auto scaling group, so if a helath check fails for an instance it will be
- * torn down and a new one deployed with the same EBS volume mounted.
+ * Elastic load balancer is created for elasticsearch API enpoint, unless ALBs are supplied
+ * as arguments. Every node is managed by an autoscaling group, so if a helath check fails
+ * for an instance it will be torn down and a new one deployed with the same EBS volume
+ * attached and mounted.
  */
 
 module "data-node-ebs-volumes" {
@@ -57,7 +58,8 @@ resource "aws_autoscaling_group" "data-node-asg" {
   launch_configuration = "${element(aws_launch_configuration.data-node-lc.*.name, count.index)}"
   health_check_type    = "ELB"
   vpc_zone_identifier  = ["${element(var.private_subnet_ids, count.index)}"]
-  target_group_arns    = ["${concat(list(aws_alb_target_group.elasticsearch-api.arn), aws_alb_target_group.elasticsearch-api-secured.*.arn)}"]
+  target_group_arns    = ["${concat(aws_alb_target_group.elasticsearch-api.*.arn, aws_alb_target_group.elasticsearch-api-secured.*.arn)}"]
+  load_balancers       = ["${concat(aws_elb.elasticsearch-internal-elb.*.name, aws_elb.elasticsearch-external-elb.*.name)}"]
 
   lifecycle = {
     create_before_destroy = true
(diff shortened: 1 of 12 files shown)


modules/elasticsearch/data-nodes.tf
```diff
@@ -3,9 +3,10 @@
  *
  * Just as with master nodes, each data node has a persistent EBS volume created for.
  * Data nodes also function as coordinators, thus each one has 9200 port open for the API.
- * Elastic load balancer is created for elasticsearch API enpoint. Every node is managed
- * by an auto scaling group, so if a helath check fails for an instance it will be
- * torn down and a new one deployed with the same EBS volume mounted.
+ * Elastic load balancer is created for elasticsearch API enpoint, unless ALBs are supplied
+ * as arguments. Every node is managed by an autoscaling group, so if a helath check fails
```
ketzacoatl: `as arguments.`  looks like it goes with the previous line.

modules/elasticsearch/data-nodes.tf
```diff
@@ -3,9 +3,10 @@
...
+ * for an instance it will be torn down and a new one deployed with the same EBS volume
+ * attached and mounted.
```
ketzacoatl: same here :point_up:

modules/elasticsearch/data-nodes.tf
```diff
@@ -191,12 +239,12 @@ resource "aws_alb_target_group" "elasticsearch-api-secured" {
...
 
 resource "aws_alb_listener" "elasticsearch-api-secured" {
-  count             = "${var.external_alb_setup ? 1 : 0}"
+  count             = "${var.external_alb_setup && !lookup(var.external_alb, "deploy_elb", false) ? 1 : 0}"
   load_balancer_arn = "${lookup(var.external_alb, "arn", "")}"
   port              = 9201
   protocol          = "HTTPS"
   ssl_policy        = "ELBSecurityPolicy-2016-08"
-  certificate_arn   = "${data.aws_acm_certificate.elasticsearch-cert.arn}"
+  certificate_arn   = "${var.external_alb["certificate"]}"
```
ketzacoatl: I would consider renaming the key in `external_alb` map to `certificate_arn`.

modules/elasticsearch/variables.tf
```diff
@@ -128,6 +128,16 @@ variable "credstash_get_cmd" {
...
+/**
+ Argument is expected to have this structure:
+  * security_group_id (required)
+  * deploy_elb (optional, default: false) - if set to true ELB will be deployed.
+  * deploy_elb_internal (optional, default: true) - if set to false ELB will be deployed as an external one.
+  * deploy_elb_cross_zone (optional, default: true) - if set to false cross AZ for ELB will be off.
+  Below are only required if `deploy_elb` is set to `false` or ommited.
+  * arn - ALB ARN
+  * dns_name - host-header for ALB listener rule
+*/
```
ketzacoatl: Would this content be better suited to the `README`?

modules/elasticsearch/variables.tf
```diff
@@ -138,11 +148,9 @@ variable "external_alb_setup" {
...
 }
 
-variable "elasticsearch_dns_ssl_name" {
-  default     = ""
-  description = "DNS name for Elasticsearch endpoint SSL. An SSL certificate is expected to be present in ACM for this domain. If left empty 'elasticsearch_dns_name' will be checked instead."
-}
-
+/* Same as `internal_alb`, except also requires a "certificate" variable, which is an
+   ARN for an ISSUED ACM certificate
+*/
```
ketzacoatl: For these couple of doc blocks in `variables.tf`, I would recommend the following:
* move the previous comment (`Argument is expected to have this structure`), and the one noted above, to each variables `description` param
* use a heredoc to allow/make use of a multi-line comment.

In doing so, the TF registry will include the info (structure of the map) in the docs it auto-generates.

In addition to that, we could include a full example in the README.

modules/kibana/variables.tf
```diff
@@ -51,6 +57,17 @@ variable "credstash_get_cmd" {
...
+ Argument is expected to have this structure:
+  * security_group_id (required)
+  * certificate (required) - ARN of ACM certificate to be used.
+  * deploy_elb (optional, default: false) - if set to true ELB will be deployed.
+  * deploy_elb_internal (optional, default: true) - if set to false ELB will be deployed as an external one.
+  * deploy_elb_cross_zone (optional, default: true) - if set to false cross AZ for ELB will be off.
+  Below are only required if `deploy_elb` is set to `false` or ommited.
+  * arn - ALB ARN
+  * dns_name - host-header for ALB listener rule
+*/
```
ketzacoatl: I would say the same here, as I did about the other block.

ketzacoatl:
I added comments on a few (minor) issues, no blockers from me :+1:  Great work @lehins!  :taco:

modules/elasticsearch/data-nodes.tf
lehins: You are absolutely right, it does go with the previous line. This is what we call wrap around :)
I am not really following what you think the problem is here, perhaps you are thinking of those as bullet points? If that's what you mean, I was just using convention that you use in other modules, where each line in a doc is prepended by `*`, which is not the same as bullet points.

modules/elasticsearch/data-nodes.tf
lehins: Same here :point_up:

modules/elasticsearch/data-nodes.tf
lehins: I was contemplating the same thing. Changed.

modules/elasticsearch/variables.tf
lehins: I was trying to think of a btter way to deal with a description for map type variables. I remember that terraform deprecated multiline description in variables, so I mistakenly assumed heredoc wouldn't work here either. Just tried it - works fine. Wondering though if it will be rendered properly in the tf registry, I guess we'll see. Changed.

As far as example goes, which README should we place it in one? `kibana`, `elasticsearch` or `elk-stack`, `elk-r53` or `web-alb module? All of them expect the same structure, so it be best to have a single place that explains it and that's where example would go too. Really wishing now terraform had types!!

modules/kibana/variables.tf
lehins: Same response

modules/elasticsearch/data-nodes.tf
ketzacoatl: LOL, you are right, I misread the `* ` as bullet points (3am amnesia).

modules/elasticsearch/variables.tf
ketzacoatl: > As far as example goes, which README should we place it in one? kibana, elasticsearch or elk-stack, elk-r53 or web-alb module? All of them expect the same structure, so it be best to have a single place that explains it and that's where example would go too.

I guess the README for the `elk-stack` module could have the one big example?

PR #75, opened 2018-03-14
Moved the hardcoded coreos ami_id to a variable from kube public and private stack

lpaulmp:
The `ami_id` changes according to the region of Amazon, this PR moved the hardcoded CoreOS ami_id to a variable.

### DIFF
--- a/examples/kube-stack-private/kube.tf
+++ b/examples/kube-stack-private/kube.tf
@@ -24,9 +24,9 @@ module "kube-cluster" {
     "${module.kube-load-balancer-sg.id}",
   ]
 
-  controller_ami        = "ami-3f061b45"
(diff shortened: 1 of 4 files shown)


ketzacoatl:
I would recommend considering defining separate variables, so the controller/worker can get separate AMIs (or the operator can control them separately). I would also consider making the variable name more generic, like `ami` or use a map as we do in other examples. Please also take note of the way I prefixed your commit message with `ex/kube-stack-*`, in this case your commit is updating the `kube-stack-public` and `kube-stack-private` example env. With a mega-module repo, these prefixes are really helpful.

My recommendations above can be addressed separately, this PR is fine to merge as-is, so I have.

Issue #57, opened 2018-02-07
all modules should add `description` to all `outputs`


ketzacoatl:
we're still using the `// description` as used by `terraform-docs`, but TF core now has support for `description` in the `output` - see https://www.terraform.io/docs/configuration/outputs.html#description-1

ketzacoatl:
@lpaulmp, This is another easy task that would show you around the modules, while also improving the auto-generated docs displayed on the upstream Terraform registry.

lpaulmp:
@ketzacoatl I opened a PR to address this issue

PR #77, opened 2018-03-16
Added descriptions of output configurations

lpaulmp:
Patch for the issue https://github.com/fpco/terraform-aws-foundation/issues/57

### DIFF
--- a/examples/gitlab-asg/main.tf
+++ b/examples/gitlab-asg/main.tf
@@ -262,22 +262,22 @@ resource "aws_route53_record" "registry" {
 ##################
 ## Outputs
 
-// region deployed to
 output "region" {
   value = "${var.region}"
+  description = " region deployed to"
 }
 
-// name of the Gitlab autoscaling group
 output "gitlab_asg_name" {
   value = "${var.name}-gitlab-asg-${element(data.aws_availability_zones.available.names, 0)}"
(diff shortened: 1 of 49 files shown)


ketzacoatl:
Minor edits to make:
* drop the leading space, there is `description = " foo"` instead of `description = "foo"`
* whitespace alignment (see `tf fmt`), eg:
 
```
variable "ssh_pubkey" {
  default     = "~/.ssh/id_rsa.pub"
  description = "file path to the pubkey to use for SSH and AWS keypair"
}
```

examples/kube-stack-private/variables.tf
```diff
@@ -18,7 +18,7 @@ variable "instance_type" {
 }
 
 variable "coreos_stable_ami_id" {
-  description = "set a stable coreos ami id"
+  description = "set a stable coreos ami id from https://goo.gl/2vJs7F"
```
ketzacoatl: I would recommend http://stable.release.core-os.net/amd64-usr/current/coreos_production_ami_all.json as a more future-proof URL over http://stable.release.core-os.net/amd64-usr/1465.7.0/coreos_production_ami_hvm.txt

lpaulmp: Sure I'm updating the link

lpaulmp:
@ketzacoatl Done can you check, please?

PR #82, opened 2018-03-23
Add Kube Controller, Worker, and Load Balancer security groups modules

lpaulmp:
Add three new security group modules, one each for the Kube Controller, Worker, and Load Balancer.

- [x] Add the modules to Kube Stack Private 
- [x] Test running the plan and apply it

### DIFF
--- a/examples/kube-stack-private/kube.tf
+++ b/examples/kube-stack-private/kube.tf
@@ -38,76 +38,20 @@ module "kube-cluster" {
   ]
 }
 
-# security group for kube controller
 module "kube-controller-sg" {
-  source      = "../../modules/security-group-base"
+  source      = "../../modules/kube-controller-sg"
   name        = "${var.name}-kube-controller"
-  vpc_id      = "${module.vpc.vpc_id}"
-  description = "Security group for the kube controllers in ${var.name}"
+  cidr_blocks = "${var.vpc_cidr}"
 }
 
-module "kube-controller-private-ssh-rule" {
-  source            = "../../modules/ssh-sg"
-  cidr_blocks       = ["${var.vpc_cidr}"]
-  security_group_id = "${module.kube-controller-sg.id}"
-}
-
-module "kube-controller-kube-api-rule" {
-  source            = "../../modules/single-port-sg"
-  port              = "6443"
-  protocol          = "tcp"
-  description       = "Allow access to kube api from hosts in ${var.name} VPC"
-  cidr_blocks       = ["${var.vpc_cidr}"]
-  security_group_id = "${module.kube-controller-sg.id}"
-}
-
-module "kube-controller-etcd-rule" {
-  source            = "../../modules/etcd-server-sg"
-  cidr_blocks       = ["${var.vpc_cidr}"]
-  security_group_id = "${module.kube-controller-sg.id}"
-}
-
-module "kube-controller-open-egress-rule" {
-  source            = "../../modules/open-egress-sg"
-  security_group_id = "${module.kube-controller-sg.id}"
-}
-
-# security group for kube worker
 module "kube-worker-sg" {
-  source      = "../../modules/security-group-base"
+  source      = "../../modules/kube-worker-sg"
   name        = "${var.name}-kube-worker"
-  vpc_id      = "${module.vpc.vpc_id}"
-  description = "Security group for the kube workers in ${var.name}"
+  cidr_blocks = "${var.vpc_cidr}"
 }
 
-# allow ingress on any port, to kube workers, from any host in the VPC
-module "kube-worker-open-ingress-rule" {
-  source            = "../../modules/open-ingress-sg"
-  cidr_blocks       = ["${var.vpc_cidr}"]
-  security_group_id = "${module.kube-worker-sg.id}"
-}
-
-module "kube-worker-open-egress-rule" {
-  source            = "../../modules/open-egress-sg"
-  security_group_id = "${module.kube-worker-sg.id}"
-}
-# security group for load balancer
-module "kube-load-balancer-sg" {
-  source      = "../../modules/security-group-base"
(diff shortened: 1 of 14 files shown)


examples/kube-stack-private/kube.tf
```diff
@@ -38,76 +38,20 @@ module "kube-cluster" {
   ]
 }
 
-# security group for kube controller
 module "kube-controller-sg" {
-  source      = "../../modules/security-group-base"
+  source      = "../../modules/kube-controller-sg"
   name        = "${var.name}-kube-controller"
```
ketzacoatl: In this case, you want to split the `${var.name}` into `${var.name_prefix}` and `${var.name_suffix}`, then give `name_suffix` a default of `kube-controller`, and set no default on `name_prefix`. That then requires the operator sets `name_prefix`, and applies a default naming scheme, but allows them to override the naming scheme if they choose.

You will see this pattern in the other modules.

examples/kube-stack-private/kube.tf
```diff
@@ -38,76 +38,20 @@ module "kube-cluster" {
...
-  vpc_id      = "${module.vpc.vpc_id}"
-  description = "Security group for the kube controllers in ${var.name}"
+  cidr_blocks = "${var.vpc_cidr}"
```
ketzacoatl: This does not look right, `cidr_blocks` would be a list.

modules/kube-controller-sg/main.tf
```diff
@@ -0,0 +1,48 @@
+/**
+ * ## Kube Controller Security Group
+ *
+ * Define a security group for kube controller
+ *
+ */
+
```
ketzacoatl: you will probably want additional variables for different CIDR blocks. For example, the list of CIDR blocks for SSH might be different for etcd, and the others.

modules/kube-controller-sg/main.tf
```diff
@@ -0,0 +1,48 @@
...
+}
+
+variable "name" {
+  description = "Name of the Kube Controller security group"
+}
+
+module "kube-controller-sg" {
+  source      = "../security-group-base"
+  name        = "${var.name}"
+  vpc_id      = "${module.vpc.vpc_id}"
```
ketzacoatl: There isn't a `vpc` module in this security group module. This should be `var.vpc_id`.

modules/kube-controller-sg/main.tf
```diff
@@ -0,0 +1,48 @@
...
+  description = "Security group for the kube controllers in ${var.name}"
+}
+
+module "kube-controller-private-ssh-rule" {
+  source            = "../ssh-sg"
+  cidr_blocks       = ["${var.vpc_cidr}"]
```
ketzacoatl: This should be its own variable.

modules/kube-controller-sg/main.tf
```diff
@@ -0,0 +1,48 @@
...
+  security_group_id = "${module.kube-controller-sg.id}"
+}
+
+module "kube-controller-kube-api-rule" {
+  source            = "../single-port-sg"
+  port              = "6443"
+  protocol          = "tcp"
+  description       = "Allow access to kube api from hosts in ${var.name} VPC"
+  cidr_blocks       = ["${var.vpc_cidr}"]
```
ketzacoatl: same here, this is a knob an operator may want to tweak independently of the list for SSH.

modules/kube-loadbalancer-sg/main.tf
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+module "kube-load-balancer-sg" {
+  source      = "../security-group-base"
+  name        = "${var.name}"
+  vpc_id      = "${module.vpc.vpc_id}"
+  description = "Security group for the kube load-balancer in ${var.name}"
+}
+
+module "kube-load-balancer-api-rule" {
```
ketzacoatl: let's rename these to `api-rule` and similar, to drop the `kube-load-balancer`. This change will make the names in the Terraform state a little shorter and easier to track.

modules/kube-loadbalancer-sg/main.tf
```diff
@@ -0,0 +1,26 @@
+/**
+ * ## Kube Load Balancer Security Group
+ *
+ * Define a security group for kube load balancer
+ *
+ */
+
+module "kube-load-balancer-sg" {
```
ketzacoatl: I'm not sure if we should rename this to `main`, `security-group`, `sg`, or `lb-sg`, but I think one of these will be better (as seen through the Terraform state).

modules/kube-worker-sg/main.tf
```diff
@@ -0,0 +1,35 @@
+/**
+ * ## Kube Worker Security Group
+ *
+ * Define a security group for kube worker
+ *
+ */
+
```
ketzacoatl: This is missing from the example your code is derived from, but it would make sense to add a rule for SSH to the workers, and the CIDR blocks variable for that.

lpaulmp:
@ketzacoatl Thanks addressing your comments

modules/kube-controller-sg/main.tf
lpaulmp: @ketzacoatl Then I would add `cidr_blocks_ssh` and `cidr_blocks_etcd` variables on the module. This variable should also be added to `kube-stack-private/variables.tf` with the same default https://github.com/fpco/terraform-aws-foundation/blob/master/examples/kube-stack-private/variables.tf#L43

modules/kube-controller-sg/main.tf
lpaulmp: @kerscher  Does this varibale sounds good `vpc_cidr_controller_api`?

ketzacoatl:
I have opened #94 to take this one over the hill (closing in favor of the new PR)

Issue #78, opened 2018-03-16
new modules: kube security groups


ketzacoatl:
Add three new security group modules, one each for the Kube Controller, Worker, and Load Balancer.

For example, for the Kube Controller, extract https://github.com/fpco/terraform-aws-foundation/blob/09af3018b4c9058544cf432de4720ec81e7fda96/examples/kube-stack-private/kube.tf#L41-L73 into a new module, parametize the details like CIDR and name (following conventions with our other modules, like variable names and such), and update the example env (in the previous link and its "public" counter-part) to use the new module.

Repeat the same process for the Kube Worker - https://github.com/fpco/terraform-aws-foundation/blob/09af3018b4c9058544cf432de4720ec81e7fda96/examples/kube-stack-private/kube.tf#L75-L93 - and the Kube Load Balancer - https://github.com/fpco/terraform-aws-foundation/blob/09af3018b4c9058544cf432de4720ec81e7fda96/examples/kube-stack-private/kube.tf#L94-L113.

ketzacoatl:
I have opened #94 to take this one over the hill (closing in favor of the new PR)

ketzacoatl:
This has been merged to master!

PR #94, opened 2018-04-13
Create SG modules for Kubernetes clusters

ketzacoatl:
This PR picks up with @lpaulmp left off in his PR for #78 (https://github.com/fpco/terraform-aws-foundation/pull/82), and primarily renames the variables and updates the use in the public/private kube-stack examples env, but also has a few more subtle changes, like ensuring that the CIDR range used by the load balancer SG is configurable through a variable (`0.0.0.0/0` is a fine default, but with a variable, a user of the module can override it). Same with the `api_port` for kubernetes load balancer and controllers.

### DIFF
--- a/examples/kube-stack-private/kube.tf
+++ b/examples/kube-stack-private/kube.tf
@@ -38,76 +38,30 @@ module "kube-cluster" {
   ]
 }
 
-# security group for kube controller
+# security group for kube controller nodes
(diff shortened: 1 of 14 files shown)


lpaulmp:
@ketzacoatl Looks good to me :shipit:, awesome work

PR #98, opened 2018-04-16
Format codebase

### LINKED ISSUE #97: Format codebase
There are multiple styles within the codebase, making it necessary for engineers to selectively commit when using the standard `terraform fmt`. We should format the entire codebase once, and then keep it as such.

Adding CI pipeline to check for it would also be a good idea, but for a second moment.

kerscher:
Fixes #97 

Since there is some counter-arguments to having a single coding style enforced by `terraform fmt`, I’m issuing the PR so we can point out directly where it’s undesired.


kerscher:
Anyone have any objections? Can we point out the problems?
@borsboom @ketzacoatl 

Otherwise we should just merge. It’s easier to adjust later to some other style if we have a consistent codebase. Right now is all over the place.

examples/vpc-scenario-1/main.tf
```diff
@@ -41,15 +41,18 @@ module "vpc" {
...
   cidr        = "10.23.0.0/16"
   azs         = ["${slice(data.aws_availability_zones.available.names, 0, 3)}"]
-  extra_tags  = { kali = "ma" }
-  public_subnet_cidrs  = ["${var.public_subnet_cidrs}"]
 
+  extra_tags = {
+    kali = "ma"
+  }
+
+  public_subnet_cidrs = ["${var.public_subnet_cidrs}"]
```
ketzacoatl: I prefer the previous, which I see as the shorter, more concise and aligned version.

examples/vpc-scenario-1/main.tf
```diff
@@ -76,18 +80,24 @@ resource "aws_instance" "web" {
...
     volume_type = "gp2"
     volume_size = "8"
   }
+
   associate_public_ip_address = "true"
-  vpc_security_group_ids      = ["${module.public-ssh-sg.id}",
-                                 "${module.open-egress-sg.id}"
+
+  vpc_security_group_ids = ["${module.public-ssh-sg.id}",
+    "${module.open-egress-sg.id}",
```
ketzacoatl: This doens't seem like much of an improvement to me, but even so.. If the previous version were to be updated, I would have opted to put the `public-ssh-sg` module on a it's own line.

examples/vpc-scenario-2-nat-instances-per-az/main.tf
```diff
@@ -81,10 +81,13 @@ module "nat-instances" {
...
   name_prefix = "${var.name}"
   key_name    = "${aws_key_pair.main.key_name}"
+
   # let AWS set IPs for us
   private_ips = []
+
   # list of subnets to deploy NAT instances into
-  public_subnet_ids    = ["${module.public-subnets.ids}"]
+  public_subnet_ids = ["${module.public-subnets.ids}"]
+
```
ketzacoatl: I find the previous version "more aligned" and preferable to lots of spaces simply b/c we have a comment to doc the param. `fmt's` default here is to leave each line's `=` to fall where it falls, and to add extra whitespace between the lines. This is counter-intuitive to me.

examples/vpc-scenario-2/main.tf
```diff
@@ -77,70 +82,80 @@ module "open-egress-sg" {
...
+  min_nodes        = "${length(module.vpc.public_subnet_ids)}"
+  public_ip        = false
+  key_name         = "${aws_key_pair.main.key_name}"
+  subnet_ids       = ["${module.vpc.private_subnet_ids}"]
+
   security_group_ids = ["${module.public-ssh-sg.id}",
-                        "${module.open-egress-sg.id}",
-                        "${aws_security_group.web-service.id}"
+    "${module.open-egress-sg.id}",
+    "${aws_security_group.web-service.id}",
```
ketzacoatl: For this last bit on security groups, I prefer the previous.

modules/aws-ipsec-vpn/main.tf
```diff
@@ -47,7 +47,7 @@ resource "aws_vpn_gateway" "main" {
 
 resource "aws_customer_gateway" "main" {
   ip_address = "${var.remote_device_ip}"
-  bgp_asn    = "65000" # required, but I don't think it's used with a static config
+  bgp_asn    = "65000"                                                     # required, but I don't think it's used with a static config
```
ketzacoatl: I don't understand why `fmt` does this with inline comments..

modules/ec2-nat-instances/main.tf
```diff
@@ -29,7 +29,8 @@ module "init-nat-hostname" {
 # this code needs to escape through (terraform, init/bash/echo, shell)
 module "init-nat-config-iptables" {
   source = "../init-snippet-exec"
-  init   = <<END_INIT
+
+  init = <<END_INIT
```
ketzacoatl: I don't see this as an improvement.

modules/ec2-nat-instances/main.tf
```diff
@@ -48,7 +49,8 @@ END_INIT
 
 # create a list of template_file data sources with init for each instance
 data "template_file" "nat-init" {
-  count    = "${length(var.public_subnet_ids)}"
+  count = "${length(var.public_subnet_ids)}"
+
```
ketzacoatl: I don't see why `fmt` would knock this out of alignment.

modules/ec2-nat-instances/main.tf
```diff
@@ -57,7 +59,8 @@ END_INIT
 }
 
 module "ec2-nat" {
-  source             = "../ec2-auto-recover-instances"
+  source = "../ec2-auto-recover-instances"
+
```
ketzacoatl: same here.

modules/ec2-nat-instances/main.tf
```diff
@@ -69,10 +72,11 @@ module "ec2-nat" {
   iam_profiles       = ["${var.iam_profiles}"]
   security_group_ids = ["${var.security_group_ids}"]
   subnet_ids         = ["${var.public_subnet_ids}"]
+
   # let AWS choose IPs for these instances
```
ketzacoatl: In the next section, removing the one extra space makes sense, but I find `fmt`'s annoying in the editor (interferes with multi-file editing).

PR #105, opened 2018-04-22
Lightweight tests

### LINKED ISSUE #103: Lightweight tests
Currently this repository has modules under `examples` that when executed will verify against AWS if modules are correct. This is the most robust and correct way to handle testing. But it suffers from a few problems:

- Running it entails costs
- If `examples` don’t cover all `modules`, then coverage is less than 100%

I propose adding a minimal test environment that can check the code _without_ running a `plan` or `apply`, by executing _at least and for every commit_:

1. `terraform init` for _all_ modules
1. `tflint`
1. checking undeclared variable types
1. checking for inline values

## `terraform init`
The easiest way to do this is to create a `tests/main.tf` file that calls every single module once, with all parameters filled with the default for each type. That means an empty string for strings; empty list for lists; and empty map for maps. `terraform init` surprisingly doesn’t care about this. It works. The same module is used for all next steps.

## `tflint`
Running against the `init`ed `tests/main.tf` finds problematic code as soon as it’s added. In my opinion we should treat `NOTICE` as OK to pass, but anything `WARNING` and above as a failed build.

## Undeclared variable types
This checks if all variables have a type at declaration site. The following is an example of a valid snippet:

```
variable "foo" {
  type        = "string"
  description = "bar"
  default     = "baz"
}
```

Omitting `type = "string"` is a test error. This guarantees that type errors are identical at every call site, as undeclared types will only be resolved at the callee, leading to potentially two modules resolving to different types. This also means that even though `default = []` will infer `type = "list"`, without adding it you will only find breakage if changing to another type once you use the changed variable at the call site.

I have started patching [`terraform-index`](github.com/kerscher/terraform-index) such that variables return their types instead of only their names. When a type hasn’t been explicitly set, it will list it as `undeclared`. Any `undeclared` should trigger a build error. This uses the `hcl` library used by `terraform` itself.

Once there are no more `undeclared`, I can use the patched `terraform-index` to auto-generate the `tests/main.tf` by using default values for every type and the list of variables for each module. The commit would be trivial.

## Inline
Inline values prevent templates from being checked as individual files externally, but this is only a nuisance. The real danger is that forgetting to close an inline delimiter changes semantics of the entire `.tf` file. Check the example below:

```
module "a" {
  source = "some-path"
  param = <<EOF
bar
EOT
}

module "b" {
  source = "some-path"
  param = <<EOF
baz
EOF
}
```

The following code above will `terraform apply` correctly, but `module.a.param` will not be `bar`, and `module.b` will never be created. In modules with nested resources or a large amount of modules this can easily pass unnoticed — specially common since `module`s take no `count`.

A counter-argument is that syntax highlight could potentially showcase this problem. But this cannot be checked without human intervention, as a checker cannot know if `module.b` should or not be created.

# Implementing

Adding under `scripts/ci/*.sh` tests that call each of the things above. Once an image has been created with the dependencies in the tests, we can add it to Travis and add a badge with build status. This last step is by far the easiest part.

kerscher:
Fixes #103


ketzacoatl:
I think this is great stuff, and I really appreciate your work on this :heart: 

My only feedback is on the topic of inlines: while I agree with you in an ideological way, I would like to roll out this change slower than immediately classifying all inlines as a problem that should be removed. I agree there is a risk to missing the closing brace, but I would prefer we review the various occurrences of this individually. For these reasons, I feel that check should not be included in the suite.

kerscher:
Nothing is running them all together for now. For #105 we can disable the inline checks if we accept the risks for automated tests :smiley: It’s OK. The current PR only gives you the ability to see it.

kerscher:
I will remove the inline checks and sync up, @ketzacoatl. We can add later :beer:

scripts/ci/tfinit.sh
```diff
@@ -0,0 +1,11 @@
+#/bin/bash
+
+set -o nounset
+set -o pipefail
+set -o errexit
```
NorfairKing: @nh2 Probably has some opinions on this part.

kerscher:
Oh, I fully agree that this should be an actual program instead, but I don’t think waiting for it to know what’s wrong with the codebase is a good idea. If anything it serves as a spec for what needs to be created. I will add an issue about this, @NorfairKing !

PR #121, opened 2018-04-27
Update the vpc-scenario-1 and vpc-scenario-2 example files. Add README's and Makefiles.

mcgirr:


examples/vpc-scenario-1/README.org
```diff
@@ -0,0 +1,40 @@
+** Example to test the VPC Scenario 1 Module
```
ketzacoatl: It's a nit-pick, but just for consistency, it would be better to use `#` for headings, as that is what we're using everywhere else

examples/vpc-scenario-1/README.org
```diff
@@ -0,0 +1,40 @@
...
+make apply
+make plan-subnets
+make apply
+make plan-gateway
+make apply
+#+END_SRC bash
+
+*** Testing
+
+Get the public IP address of the vpc machine with the AWS console.
```
ketzacoatl: Does "vpc machine" refer to the web instance? This terminology is a little unclear.

examples/vpc-scenario-1/main.tf
```diff
@@ -49,29 +63,26 @@ module "vpc" {
...
+  security_group_id = "${module.vpc-sg.id}"
 }
 
 # shared security group, open egress (outbound from nodes)
 module "open-egress-sg" {
   source = "../../modules/open-egress-sg"
-  name   = "${var.name}"
-  vpc_id = "${module.vpc.vpc_id}"
+
+  # vpc_id = "${module.vpc.vpc_id}"
```
ketzacoatl: drop the commented line.

examples/vpc-scenario-1/README.org
mcgirr: Should I be using markdown instead of Org?

mcgirr: I can rewrite the README file as a markdown file instead of an Org file.

examples/vpc-scenario-2/main.tf
```diff
@@ -207,7 +219,13 @@ END_INIT
...
+
 output "elb_dns" {
   value       = "${aws_elb.web.dns_name}"
   description = "make the ELB accessible on the outside"
 }
+
+resource "aws_key_pair" "main" {
+  key_name   = "${var.name}"
+  public_key = "${file(var.ssh_pubkey)}"
+}
```
ketzacoatl: I don't see a reason why this resource should be moved to the bottom. Without a purpose we can articulate, this type of churn is discouraged.

examples/vpc-scenario-2/main.tf
```diff
@@ -143,7 +155,7 @@ module "web" {
   source           = "../../modules/asg"
   ami              = "${module.ubuntu-xenial-ami.id}"
   azs              = "${slice(data.aws_availability_zones.available.names, 0, 3)}"
-  name             = "${var.name}-web"
+  name_prefix      = "${var.name}"
```
ketzacoatl: I believe the `-web` should be kept.

examples/vpc-scenario-2/main.tf
```diff
@@ -44,7 +54,7 @@ module "vpc" {
   source      = "../../modules/vpc-scenario-2"
   name_prefix = "${var.name}"
   region      = "${var.region}"
-  cidr        = "10.23.0.0/16"
+  cidr        = "${var.vpc_cidr}"
```
ketzacoatl: While minor, this is great to see, thanks!

examples/vpc-scenario-2/main.tf
```diff
@@ -14,6 +19,11 @@ variable "region" {
   default     = "us-east-2"
```
ketzacoatl: Might as well make `region` a variable while adding `vpc_cidr`.

examples/vpc-scenario-1/main.tf
```diff
@@ -49,34 +63,30 @@ module "vpc" {
...
-  name   = "${var.name}"
-  vpc_id = "${module.vpc.vpc_id}"
+
+  security_group_id = "${module.vpc-sg.id}"
 }
 
 resource "aws_instance" "web" {
   ami               = "${module.ubuntu-xenial-ami.id}"
-  count             = "${length(module.vpc.public_subnet_ids)}"
+  count             = "1"
```
ketzacoatl: why this change?

examples/vpc-scenario-2/main.tf
```diff
@@ -60,24 +70,25 @@ module "ubuntu-xenial-ami" {
...
   source              = "../../modules/ssh-sg"
-  name                = "${var.name}"
-  vpc_id              = "${module.vpc.vpc_id}"
-  allowed_cidr_blocks = "0.0.0.0/0"
+
+  security_group_id = "${module.vpc-sg.id}"
 }
 
 # shared security group, open egress (outbound from nodes)
 module "open-egress-sg" {
```
ketzacoatl: I don't think this is happening elsewhere, but we should probably consider updating our scheme to include `rule` in the modules that are adding rules. We would then rename these modules from `open-egress-sg` to `open-egress-rule` or `open-egress-sg-rule`. I'm not sure if it makes sense to do that here and now, or wait and do that across the board all in one fell swoop.

The context here: we used to provide a set of modules that created their own security group, which meant attaching multiple SG to a single instance. There's a limit on the number of SG you can attach (6, I think), so that eventually didn't work. Also, TF added support for "creating a SG, then attached N rules to the SG", and we decided to refactor our modules around the concept.. so in the initial 0.7.x release, we refactored the SG modules to provide a "base SG" and then other modules to "attach rules" to that base SG.

examples/vpc-scenario-2/main.tf
```diff
@@ -60,24 +70,25 @@ module "ubuntu-xenial-ami" {
   release = "14.04"
 }
 
-resource "aws_key_pair" "main" {
-  key_name   = "${var.name}"
-  public_key = "${file(var.ssh_pubkey)}"
+module "vpc-sg" {
```
ketzacoatl: We usually name the SG based on the primary role of the hosts the SG is attached to. So if we have an ELB with a SG, the name of the SG will have elb in it. Here is my recommendation: create 2 SG with the base SG module, one for ELB and one for the web instance, and then attach rules to each of those SG to match what was in place before.

examples/vpc-scenario-2/main.tf
```diff
@@ -60,24 +70,25 @@ module "ubuntu-xenial-ami" {
...
+  source      = "../../modules/security-group-base"
+  description = "Test project security group"
+  name        = "${var.name}-vpc-sg"
```
ketzacoatl: drop the `-sg` from the name - in the UI when we see this particular name, we'll know it's a Security Group.

examples/vpc-scenario-2/main.tf
```diff
@@ -60,24 +70,25 @@ module "ubuntu-xenial-ami" {
...
 }
 
 # shared security group for SSH
 module "public-ssh-sg" {
   source              = "../../modules/ssh-sg"
-  name                = "${var.name}"
-  vpc_id              = "${module.vpc.vpc_id}"
-  allowed_cidr_blocks = "0.0.0.0/0"
+
+  security_group_id = "${module.vpc-sg.id}"
```
ketzacoatl: drop the extra newline and align the `=`. Can also use `terraform fmt` to help with alignment and formatting issues.

examples/vpc-scenario-1/main.tf
mcgirr: @ketzacoatl  It was one way I found to resolve an error that I was getting. 
When count is set to `${length(module.vpc.public_subnet_ids)}` instead of 1 the following error is produced:

```
Error: Error applying plan:

2 error(s) occurred:

* aws_instance.web[1]: 1 error(s) occurred:

* aws_instance.web.1: Error launching source instance: InvalidParameterValue: Value (us-east-2a) for parameter availabilityZone is invalid. Subnet 'subnet-439b4139' is in the availability zone us-east-2b
        status code: 400, request id: 0deb09d8-1f53-40ab-a763-5fbda58905a3

* aws_instance.web[2]: 1 error(s) occurred:

* aws_instance.web.2: Error launching source instance: InvalidParameterValue: Value (us-east-2a) for parameter availabilityZone is invalid. Subnet 'subnet-73b9b13e' is in the availability zone us-east-2c
        status code: 400, request id: 9b6acab4-e5b5-4880-bdbc-3d5f3904a290
```

There is probably a better way to address this error though.

examples/vpc-scenario-2/main.tf
mcgirr: So do you mean that instead of having a separate region variable, like it currently it, that `region` should be nested and set within the `vpc_cidr` variable?

examples/vpc-scenario-2/main.tf
mcgirr: @ketzacoatl The "vpc-sg" module's ID is used by the "public-ssh-sg" and the "open-egress-sg" as their `security_group_id`.  Should I use the newly created security group for ELB or the security group for the web instance as the `security_group_id` (since the "vpc-sg" module will not be there)?

examples/vpc-scenario-2/main.tf
ketzacoatl: No, that comment of mine was confused. I think I mistakenly thought I saw the AWS provider using a hardcoded region, but I don't see that and am not sure how I got off track. Sorry for the noise!

examples/vpc-scenario-2/main.tf
mcgirr: @ketzacoatl I think this relates to what you mentioned here. Since there is no longer a `vpc-sg` module and no longer a `vpc-sg.id` value to set the `security_group_id` to for the open-egress-sg and public-ssh-sg modules, in my next commit I set `security_group_id` to `vpc-elb-sg.id`.

ketzacoatl:
Rather than put you through another round going back and forth, I checked out the WIP branch, rebased to current `master`, and then squashed, updated, and re-ran through the commits to create my own version of your branch. The result can be seen here: https://github.com/fpco/terraform-aws-foundation/pull/127

mcgirr:
I am closing this PR since the changes were handled in #127

PR #127, opened 2018-05-07
Update  VPC Scenario 1 and 2 env for consistency

ketzacoatl:
This is primarily @mcgirr's work, with only minor fixups from me, and published on this PR for comparison against the existing PR (https://github.com/fpco/terraform-aws-foundation/pull/121). Whether we use Mike's original PR or this does not matter to me, I primarily wanted to give @mcgirr an easy way to see my updates and compare the differences. Either update your PR or close out and use this. See also how broke up the commits and also formatted the commit messages for readability. I consider these best practices on a repo like this one.

### DIFF
--- a/examples/vpc-scenario-2/main.tf
+++ b/examples/vpc-scenario-2/main.tf
@@ -4,6 +4,11 @@
  *
  */
 
+variable "extra_tags" {
+  description = "Extra tags that will be added to aws_subnet resources"
+  default     = {}
+}
+
 variable "name" {
   description = "name of the project, use as prefix to names of resources created"
   default     = "test-project"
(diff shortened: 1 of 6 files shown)


examples/vpc-scenario-2/main.tf
```diff
@@ -188,25 +188,6 @@ docker run                   \
 END_INIT
 }
 
-#resource "aws_instance" "bastion" {
```
mcgirr: @ketzacoatl This section (L191 - L209) was commented out so I felt okay deleting it. I saw that you left it out as well in your commits. I figured it was residual work from whoever originally wrote the example.

Issue #99, opened 2018-04-16
Update documentation in s3-remote-state


AronT-TLV:
Current documentation is way out of date and just plain wrong. Needs to be updated.

ketzacoatl:
This is complete!

PR #130, opened 2018-05-08
mod/s3-remote: #99 update README to work with latest versions of S3

AronT-TLV:
## Purpose
To address issue #99 

## Work Done
* changed main.tf to include latest TF approach
* explained how to use principals
* gave detailed up to date explanation of how to properly use partial configuration

## Achieved Goal
A more thorough and updated documentation of how to use the module for remote TF state on S3.

@ketzacoatl

### DIFF
--- a/modules/s3-remote-state/README.md
+++ b/modules/s3-remote-state/README.md
@@ -1,16 +1,18 @@
 # Remote State on S3
 
-This module simplifies the setup of an S3 bucket for Terraform's remote state
-storage. The module will:
+## Purpose of This Module
 
-* create the S3 bucket with versioning enabled
-* create an IAM policy that has full access to the bucket
-* attach a list of users and/or groups to that policy
+The purpose of this module is to create a remote backend to store Terraform deployment state in an S3 bucket. The module:
 
+* creates the S3 bucket with versioning enabled
+* creates an IAM policy that has full access to the bucket
+* works with a specific role you define
 
-## How to use this module
+## How to Use This Module
 
-### Minimal Use
+The use of this module has two stages. In the first stage we use a wrapper project to create the S3 bucket with appropriate policies. In the second stage we configure and initialize the remote backend. At the end of that process the state will be copied over to the remote backend and be available for other people to use as well.
+
+### Wrapper Project
 
 Create a new Terraform project and a `main.tf` with the following:
 
@@ -25,10 +27,16 @@ variable "principals" {
 
 provider "aws" { }
 
+#terraform {
+#  backend "s3" {
+#    encrypt = "true"
+#    bucket  = "fpco-tf-remote-state-test"
+#    key     = "mod/remote-state-storage/terraform.tfstate"
+#  }
+#}
+
 module "s3-remote-state-bucket" {
-    source      = "../../../../fpco-terraform-aws/tf-modules/s3-remote-state"
-    iam_users   = "${var.iam_users}"
-    iam_groups  = "${var.iam_groups}"
+    source      = "fpco/foundation/aws//modules/s3-remote-state"
     bucket_name = "${var.bucket_name}"
 }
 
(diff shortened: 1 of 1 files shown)


modules/s3-remote-state/README.md
```diff
@@ -1,16 +1,18 @@
...
+* creates the S3 bucket with versioning enabled
+* creates an IAM policy that has full access to the bucket
+* attaches a list of users and/or groups to that policy
 
 
-## How to use this module
+## How to Use This Module
 
-### Minimal Use
+The use of this module has two stages. In the first stage we use a wrapper project to create the S3 bucket with appropriate policies. In the second stage we configure and initialize the remote backend. At the end of that process the state will be copied over to the remote backend and be available for other people to use as well. F
```
ketzacoatl: Hanging "F".

modules/s3-remote-state/README.md
```diff
@@ -25,8 +27,16 @@ variable "principals" {
 
 provider "aws" { }
 
+#terraform {
+#  backend "s3" {
+#    encrypt = "true"
+#    bucket  = "fpco-tf-remote-state-test"
+#    key     = "mod/s3-remote/terraform.tfstate"
```
ketzacoatl: I'd recommend a `key` set to `remote-state/terraform.tfstate`, or `remote-state-storage`, or `remote-state-store`, or similar.

modules/s3-remote-state/README.md
```diff
@@ -37,43 +47,80 @@ output "bucket_name" {
...
-Push the existing state to that bucket:
+Initializing the backend...
+bucket
+  The name of the S3 bucket
 
-```
-ᐅ tf remote push
-State successfully pushed!
-```
+  Enter a value: fitstation-tf-remote-state-test
```
ketzacoatl: fixup here.

modules/s3-remote-state/README.md
```diff
@@ -37,43 +47,80 @@ output "bucket_name" {
...
-Review the changes Terraform will make with `tf plan`, and apply those changes
-with `tf apply`.
+Initially the project with `terraform init.` Review the changes Terraform will make with `tf plan --out=tf.out`, and apply those changes with `tf apply tf.out`.
+
+### Initialize the Remote State S3 Backend
 
-Configure remote state to use that bucket with:
+We want to use the newly created S3 bucket to store the Terraform state for this module. There are several steps that need to be taken. Make sure to backup your existing `terraform.tfstate` file before taking these steps:
 
+1. First we have to [initialize the backend][1] and copy over the current state of the backend bucket you just created. Uncomment the following, so that the code in the `main.tf` file looks similar to this:
```
ketzacoatl: I'm not sure why we should recommend using this partial config method when the commented out keys a) have all that is needed, and b) end up with a simpler/clearer/more explicit result, or am I overlooking something that makes the partial config work better?

modules/s3-remote-state/README.md
```diff
@@ -37,43 +47,80 @@ output "bucket_name" {
...
+
+To prevent automatic upgrades to new major versions that may contain breaking
+changes, it is recommended to add version = "..." constraints to the
+corresponding provider blocks in configuration, with the constraint strings
+suggested below.
+
+* provider.aws: version = "~> 1.14"
+
+Terraform has been successfully initialized!
+```
```
ketzacoatl: add an empty line between the closing quotes and the next line of text.

modules/s3-remote-state/README.md
AronT-TLV: The partial config method is 1/ recommended in TF doco and 2/ is only for the initial set up of the s3 project itself. It allows you to switch the s3 local project over to remote safely. As noted in point 3, when other people create projects, they should use the uncommented version so they will start with remote from the get go. I think the wording there is confusing and perhaps I should make that clearer and more explicit.

modules/s3-remote-state/README.md
AronT-TLV: fixed

modules/s3-remote-state/README.md
AronT-TLV: fixed

modules/s3-remote-state/README.md
AronT-TLV: fixed

AronT-TLV:
@ketzacoatl fixed according to your comments above

ketzacoatl:
This LGTM! I've rebased and squashed it then merged to `master`..

Issue #135, opened 2018-05-14
BIND module fails for empty `db_records_folder` parameter


kerscher:
# Expected behaviour

1. User instantiates `bind-server` and provides empty `db_records_folder`
1. The internal `null_resource` detects the empty parameter and skips copying any folder

# Actual behaviour

1. `"${var.db_records_folder}/"` is interpolated into `/`
1. Calling user sends `/` to `/tmp/db_records`
1. Permission failure to access all files under `/` fails the provisioner
1. Local `/` files are available on the instance
1. `terraform apply` fails

## Proposed solution

Send a dummy folder that gets removed upon `remote_exec`.

ketzacoatl:
I believe these changes have been merged, though we haven't seen them run recently. I am going to close the issue as complete and open new tickets in the future if there are still issues here.

PR #136, opened 2018-05-14
bind-server: fix empty `db_records_folder` file provisioner, #135

kerscher:
If `${var.db_records_folder}` was empty, this module would try to copy
`/` from the calling machine to `/tmp/db_records`. Besides being a
clear security concern, this also means it will fail, as the calling
user doesn’t have access to the entire root tree (shouldn’t at
least!).

The fix is to send a templated folder with nothing upstream.
`remote_exec` then does the right thing™.

### DIFF
--- a/modules/bind-server/main.tf
+++ b/modules/bind-server/main.tf
@@ -98,6 +98,7 @@ resource "aws_instance" "bind" {
 
   provisioner "remote-exec" {
     connection {
+      type                = "ssh"
       host                = "${self.private_ip}"
       user                = "${var.distro == "ubuntu" ? "ubuntu" : "ec2-user"}"
       private_key         = "${file(var.ssh_key)}"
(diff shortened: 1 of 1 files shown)


modules/bind-server/main.tf
```diff
@@ -97,6 +97,8 @@ resource "aws_instance" "bind" {
 
   provisioner "remote-exec" {
     connection {
+      type                = "ssh"
+      agent               = true
```
ketzacoatl: It looks like this is disabled by default - https://www.terraform.io/docs/provisioners/connection.html#agent - will enabling `agent` lead to failure if `ssh-agent` is not available/setup in the host/environment running Terraform?

kerscher: If you don’t have an agent, you wouldn’t be able to forward the key for the next hop with the bastion anyway, isn’t that correct?

ketzacoatl: Is a bastion required? What about when you aren't using one?

kerscher: You are right! Removed `agent = true`.

kerscher:
Rebased with latest `master` after applying review changes. Should be good to merge now.

PR #148, opened 2018-05-29
init-snippet-curator: make master_only option configurable

giancarlocosta:
See: https://www.elastic.co/guide/en/elasticsearch/client/curator/current/configfile.html#master_only


ketzacoatl:
is this missing the update to the template itself?

giancarlocosta:
Nope. 😃 The template allows for it but the vars passed into the template were hardcoded or some reason.

PR #149, opened 2018-06-04
`kube-stack`: attach ASG to ELB

kerscher:
Three commits, so actual change can be seen isolated.

### DIFF
--- a/modules/kube-stack/load-balancer.tf
+++ b/modules/kube-stack/load-balancer.tf
@@ -10,6 +10,7 @@ resource "aws_elb" "kube-controllers" {
     lb_port           = 443
     lb_protocol       = "tcp"
   }
+
   health_check {
     healthy_threshold   = 2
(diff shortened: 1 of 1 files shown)


modules/kube-stack/load-balancer.tf
```diff
@@ -4,32 +4,32 @@ resource "aws_elb" "kube-controllers" {
...
   internal        = "${var.private_load_balancer}"
 
-  #listener {
-  #  instance_port = 6443
-  #  instance_protocol = "https"
-  #  lb_port = 8080
-  #  lb_protocol = "http"
-  #  #ssl_certificate_id = "${data.aws_acm_certificate.kube-cert.arn}"
-  #}
-
```
ketzacoatl: I'm fine with removing this section, but maybe you could also add a github ticket to hook up TLS to the ELB in the future?

Issue #61, opened 2018-02-28
example/gitlab: swap ELB for EIP


ketzacoatl:
in the single-node asg gitlab example, we currently use an ELB. It would be nice to use EIP there instead, and demo that proper setup. An ELB is better for a configuration where we have multiple web workers in the gitlab ASG (and shared data, so no single-node asg). Here is the example env to update: https://github.com/fpco/terraform-aws-foundation/tree/master/examples/gitlab-asg

Given the non-working state of this example, split up into various phases:
- [x]  Get to a state where it's mostly working, documented and as automated as possible
- [x] A persistent EBS volume is mounted to the instance and all gitlab data is stored on that volume - data is retained when the node is replaced
- [ ] Add SMTP server for password notifications
- [ ] Instead of an ELB we use a fixed IP (original task)

ketzacoatl:
This is a good task to get acquainted with Terraform on AWS and the workflows that we are using.

AronT-TLV:
If I understand correctly, the purpose of a single node ASG is to always guarantee there is one node "alive". However to make sure your DNS works if the single node fails and automatically gets replaced, you still need the ELB, so you can CNAME to that which doesn't go away. And AFAIK you still can't use EIP with ELB. I checked as much as I can and even with the new application load balancers, it seems it's still CNAME based--unless I am missing something.

So it seems the only way to use EIP is with a single EC2 instance where you don't care if it goes down. Is that really an important use case? (This is why Kubernetes is so nice).

In any case, if it is useful, is it ok if I rephrase this task to: "create an *additional* example with a gitlab on a single EC2 instance with an EIP attached"? Because I think the current example is still useful.

ketzacoatl:
In the "single node asg" pattern, we define an auto scaling group, but limit it to 1 node. The 1 node can then have a few special resources associated with it, like an EBS volume, or an EIP. If the node is lost, a new one comes up in its place. Recovery is relatively quick, and can be automated in a variety of ways. This workflow also plays well for various upgrade/maintenance/migration scenarios. When the ASG's node comes online, it has the init code it needs to "mount" those "special resources", attaching its EBS volume or EIP. This pattern is used quite widely, as it makes for a great way to host services like gitlab (with a bunch of data), and it also provides the foundation for a "leader" or "controller" role within a cluster (like in kubernetes, controllers deployed by kops are using the Single-Node ASG pattern).

[...]

PR #154, opened 2018-06-12
61-example/gitlab asg part 1

AronT-TLV:
## Purpose
To address issue #61 first part

## Work done
- Updated the Makefile so all steps properly work
- Automated several steps using `jl`
- Clearly documented all steps
- List remaining to-dos for this issue

## Achieved Goal
Given the state of the example, it was necessary to make several changes to get it to work at all. The opportunity was used to improve documention and automate steps where possible making it clearer and easier to use.

@ketzacoatl


examples/gitlab-asg/Makefile
```diff
@@ -61,23 +60,61 @@ plan:
...
+
 ## use ops cli tool to lookup the public IP of the gitlab instance
-get-ip:
-	@ops aws ec2 asg ips -r $$(terraform output region) $$(terraform output gitlab_asg_name)
+##get-ip:
+##	@ops aws ec2 asg ips -r $$(terraform output region) $$(terraform output gitlab_asg_name)
+
+## SSH to the gitlab server and use parted and similar tools to setup a partition and filesystem on the EBS volume (run only once per volume)
+init-ebs-volume:
+	@ssh data-ops-gitlab "sudo parted --script /dev/xvdf -- mklabel msdos && \
```
borsboom: Since this target doesn't currently work, I think performing these commands on the instance manually should be documented in the README.

examples/gitlab-asg/Makefile
```diff
@@ -40,8 +35,12 @@ install-cfssl:
 
 ## Use AWS cli tools to upload the SSL cert to AWS
 upload-tls-certs:
-	@AWS_DEFAULT_REGION=us-east-1 aws acm import-certificate --certificate file://gitlab.pem --private-key file://gitlab-key.pem --certificate-chain file://ca.pem
-
+	@. ./vars.env && envsubst < upload-gen-cert.tpl > upload-gen-cert.sh
+	@. ./vars.env && envsubst < delete-gen-cert.tpl > delete-gen-cert.sh
+	@chmod +x ./upload-gen-cert.sh ./delete-gen-cert.sh
+	@./upload-gen-cert.sh
+	@echo "ssl_arn = " `cat  upload-gen-cert-arn.txt` >> terraform.tfvars
```
borsboom: Appending to `terraform.tfvars` here assumes that `make render-tfvars` has already been run, and also means that re-running `make render-tfvars` later will "erase" the value written by this step (so it has to be re-run).  Perhaps better would be to make this file an _input_ to `make render-tfvars`, and move the step of running that command to be after `make upload-tls-certs`.

examples/gitlab-asg/Makefile
```diff
@@ -5,29 +5,24 @@
...
 ## Use ssh-keygen to generate a keypair for SSH
 generate-ssh-key:
 	@ssh-keygen -q -N "" -b 4096 -C "test" -f ./id_rsa
 
 ## write out a tfvars file for a test deploy
 generate-tfvars:
-	@echo "ssh_pubkey = \"./id_rsa.pub\"" > terraform.tfvars
-	@echo "region = \"us-east-2\"" >> terraform.tfvars
-
+	@echo "ssh_pubkey = \"./id_rsa.pub\"" >> terraform.tfvars
```
borsboom: Any reason this can't be part of `make render-tfvars`, and then change the README to have you run `make render-tfvars` later in the process (e.g. just before `make network`)?  I don't think any of the earlier steps depend on `terraform.tfvars` existing.

examples/gitlab-asg/Makefile
AronT-TLV: I think a better approach is to remove them from the Makefile at this point. They aren't currently in Master so it's not a regression. The whole point is to get in a bunch of fixes that have been sitting around for a very long time into `master`. I will save these commands and come back to them in next one or two MRs.

examples/gitlab-asg/Makefile
AronT-TLV: Great catch! These two steps were added later and layered on and I didnt notice you could easily move the `render_tfvars` and consolidate with the other two modifications.

Fixed and tested.

examples/gitlab-asg/Makefile
AronT-TLV: Fixed as per above

examples/gitlab-asg/Makefile
AronT-TLV: Fixed along with next comment.

examples/gitlab-asg/Makefile
borsboom: I do think it would be good to document the need to run these commands in the README.  Otherwise, you lose all your data if the instance is ever replaced by the ASG (which is, after all, the whole point of having the ASG).

AronT-TLV:
@borsboom made change to README as per your request

ketzacoatl:
Rather than push over this PR, I have created a new PR based on this one - see #159 for that. The primary reason for a new PR was to extract the commits from the various branch merges and squash them down to something smaller in the history. The new branch is nearly identical with the old one save for:

```
diff --git a/tests/gitlab-asg/cert-config.json b/tests/gitlab-asg/cert-config.json
deleted file mode 100644
index f8a9902..0000000
--- a/tests/gitlab-asg/cert-config.json
+++ /dev/null
@@ -1,13 +0,0 @@
-{
-  "signing": {
-    "default": {
-      "expiry": "8760h",
-      "usages": [
-        "signing",
-        "key encipherment",
-        "server auth",
-        "client auth"
-      ]
-    }
-  }
-}
```

PR #155, opened 2018-06-14
asg: remove 'desired_capacity' and add 'termination_policies' input

borsboom:
When the capacity of an ASG should "float" (e.g. when using autoscaling
policies or schedules), setting the 'desired_capacity' in the ASG is
undesirable as it means anytime the TF is applied the capacity gets
reset.

Unfortunately it is not possible to make 'desired_capacity' optional, so
opted for removing it entirely.  'min_size' and 'max_size' can be used to
control the size instead.

Also added 'termination_policies' input, which is useful when autoscaling
to make sure the right instances are terminated when scaling down.

This change also required updating other modules that use the 'asg' module,
thus also removing desired capacity inputs from them.


ketzacoatl:
@borsboom, should we deploy the gitlab example env, or kube-stack-private, or similar, to test this update, or were you able to confirm the example env are still green?

borsboom:
I've used the updated module in a client's environment, so I know it works in that sense, but I haven't tried any of the examples to confirm that no incompatibilities were introduced.

ketzacoatl:
OK, thanks for the info. I'll run a couple of deploys today for a quick test. Also, for other reasons, I cut a `0.7.6` since this PR came in, so I updated rebased this branch to sync with the CHANGELOG updates. I think this is otherwise good to merge.

One last question, it would be really awesome to get this (use of the scaling policies) into an example, would you be interested in adding that separately, or maybe share a gist that one of us could use to get an example in place?

PR #164, opened 2018-07-26
gitlab-asg: enhancements

neilmayhew:
Some or all of these commits can be squashed before merging


examples/gitlab-asg/Makefile
```diff
@@ -58,45 +91,46 @@ plan:
...
+	 envsubst <ssh_config.tpl >ssh_config
 
 ## use ops cli tool to lookup the public IP of the gitlab instance
 ##get-ip:
 ##	@ops aws ec2 asg ips -r $$(terraform output region) $$(terraform output gitlab_asg_name)
 
 ## SSH to the gitlab server and use parted and similar tools to setup a partition and filesystem on the EBS volume (run only once per volume)
 init-ebs-volume:
-	@ssh data-ops-gitlab "sudo parted --script /dev/xvdf -- mklabel msdos && \
+	@ssh -F ssh_config data-ops-gitlab "sudo parted --script /dev/xvdf -- mklabel msdos && \
```
ketzacoatl: is the `data-ops-gitlab` bit a hardcoded relic from my previous work? Does that need to be parametized somewhere?

examples/gitlab-asg/Makefile
```diff
@@ -1,44 +1,77 @@
...
 	@echo "ca-key.pem is the file to keep very safe"
 
 ## use curl to download and install the cfssl executable
 install-cfssl:
-	@curl -s -L -o ./cfssl https://pkg.cfssl.org/R1.2/cfssl_linux-amd64
-	@curl -s -L -o ./cfssljson https://pkg.cfssl.org/R1.2/cfssljson_linux-amd64
-	@chmod +x ./cfssl ./cfssljson
-	@./cfssl version
+	@mkdir -p bin
+	@curl -s -L -o bin/cfssl https://pkg.cfssl.org/R1.2/cfssl_linux-amd64
```
ketzacoatl: why use `./bin` over `~/bin` ?

examples/gitlab-asg/Makefile
neilmayhew: That name is now purely local, because of using a custom `ssh_config` instead of putting the entries into `~/.ssh/config`. So the name is arbitrary, and it actually makes things easier if it's always the same.

examples/gitlab-asg/Makefile
neilmayhew: I wanted to keep everything local and not touch the user's global config at all. So, in particular, if there's already a `cfssl` in `~/bin` we shouldn't be stomping on it. We remove `./bin` at the end, in `clean`, so it should be a case of "We Were Never Here".

Also, I think the default should be to install `cfssl` from a package, so this should be used only by someone who's in a hurry and just wants to try things out quickly, which is another reason we shouldn't be leaving stuff behind in the global env.

I did consider making the download conditional on whether we can detect `cfssl` already being on the `PATH` somewhere (eg using `which` or `type`).

ketzacoatl:
Great work, thanks @neilmayhew !

Issue #169, opened 2018-08-05
replace self-signed certificate in `gitlab-ha` with AWS certificate from Certificat Manager


AronT-TLV:
AWS provides facility for easily adding a signed certificate to an ELB. This is easier, cleaner solution than the current procedure. Moreover it attaches a certificate all browsers will accept without forcing them to.

PR #170, opened 2018-08-07
ex/gitlab-ha: use AWS CM for TLS instead of self-signed cert

AronT-TLV:
Revised and simplified example to make use of AWS Certificate Manager in place of self-signed certificate -- issue #169


examples/gitlab-ha/README.md
```diff
@@ -35,17 +30,14 @@ First, copy `vars.mk.sample` to `vars.mk` and review/update the variables define
 
 * `DNS_ZONE_NAME` to the one associated with your AWS account
 * `REGION` to one that has some resources available to you
-* `REGISTRY_S3_BUCKET_NAME` to something unique
```
ketzacoatl: Did this one get dropped somewhere along the line?

examples/gitlab-ha/vars.mk.sample
```diff
@@ -2,13 +2,8 @@ export DNS_ZONE_NAME=example.net
...
-export CA_NAME=Example Net CA
-export ORGANIZATION=Example Net
-export COUNTRY=US
-export STATE=CA
-export LOCALITY=SF
+export GITLAB_REGISTRY_NAME=registry
 export GITLAB_URL=${GITLAB_NAME}.${DNS_ZONE_NAME}
-export REGISTRY_URL=${REGISTRY_NAME}.${DNS_ZONE_NAME}
+export REGISTRY_URL=${GITLAB_REGISTRY_NAME}.${DNS_ZONE_NAME}
 export REGISTRY_S3_BUCKET_NAME=abc123def456
```
ketzacoatl: is this not used and cruft from the past?

examples/gitlab-ha/README.md
AronT-TLV: mistake. will push fix

examples/gitlab-ha/vars.mk.sample
AronT-TLV: its actually a buglet and causes problems because it's named this way in the module. I fixed it in the simple-ha and didn't notice the problem here, so took the opportunity to fix it now.

Issue #158, opened 2018-07-19
update modules to use the AWS partition datasource


ketzacoatl:
Long, long ago, this didn't exist (at least I don't believe it did..) and we needed a way to deploy to GovCloud, so we put in the `aws_cloud` variable. We now have this wonderful datasource available which will provide that info for us, see https://www.terraform.io/docs/providers/aws/d/partition.html for more info.

ketzacoatl:
Wow, that was a lot to update, thanks for that!

PR #173, opened 2018-08-17
Add the AWS partition datasource for issue #158

mcgirr:
This should be able to close issue #158 

This uses the `aws_partition` data source in place of the the `aws_cloud` variable.

I believe this commit is comprehensive but there may be one that I missed.

### DIFF
--- a/modules/consul-demo-server/main.tf
+++ b/modules/consul-demo-server/main.tf
@@ -12,7 +12,6 @@ module "consul-server" {
   source                  = "../single-node-asg"
   name_prefix             = "${var.name_prefix}"
   name_suffix             = "consul-server"
-  aws_cloud               = "${var.aws_cloud}"
   region                  = "${var.region}"
   root_volume_type        = "gp2"
   root_volume_size        = "8"
--- a/modules/ec2-nat-instances/main.tf
+++ b/modules/ec2-nat-instances/main.tf
@@ -15,7 +15,6 @@
 module "ubuntu-ami" {
   source      = "../ami-ubuntu"
   release     = "16.04"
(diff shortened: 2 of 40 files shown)


modules/consul-demo-server/main.tf
```diff
@@ -8,11 +8,14 @@
...
 
+# Lookup the current AWS partition
+data "aws_partition" "current" {}
+
 module "consul-server" {
   source                  = "../single-node-asg"
   name_prefix             = "${var.name_prefix}"
   name_suffix             = "consul-server"
-  aws_cloud               = "${var.aws_cloud}"
+  aws_cloud               = "${data.aws_partition.current.partition}"
```
ketzacoatl: If you drop this variable in the other modules, it won't be used here.

modules/ec2-nat-instances/main.tf
```diff
@@ -15,7 +15,7 @@
 module "ubuntu-ami" {
   source      = "../ami-ubuntu"
   release     = "16.04"
-  is_govcloud = "${var.aws_cloud == "aws-us-gov" ? "true" : "false"}" # Canonical
+  is_govcloud = "${data.aws_partition.current.partition == "aws-us-gov" ? "true" : "false"}" # Canonical
```
ketzacoatl: this is the other variable that would be removed.

modules/consul-demo-server/main.tf
mcgirr: So, just for clarity, what you're saying is that because no other modules should be using `aws_cloud` anymore, it's okay then to remove the `aws_cloud` variable that is declared here?

Issue #152, opened 2018-06-07
new example: terraform VPC for kops


ketzacoatl:
Create a new example that demonstrates how to deploy kubernetes with `kops` using our workflow based on Terraform.

PR #177, opened 2018-08-28
example: kops in vpc made by terraform

aniketd:
Resolves: #152 

- [x] add `delete state bucket` for `kops` in the Makefile
- [x] add the README
- [x] more factoring in ENV files if possible
- [x] push corrections to PR in docs
- [x] address review comments
- [x] fix cross-copying of `env` variables
- [x] replace `kops-state-bucket` with `s3-remote-state` module
- [x] fix clean up of `*.pem*` files
- [x] word the part about using the example `env` files in the README better
- [x] move copy commands into Makefile
- [x] move `jq` command into Makefile (was easy with `$$` ;)


examples/kops-kube-vpc/Makefile
```diff
@@ -0,0 +1,80 @@
...
+
+## Check if tf.out is stale (Older than 2 minutes)
+check-plan-file:
+	@if ! ${STALE_PLAN_FILE} ; then \
+		echo "ERROR: Stale tf.out plan file (older than 3 minutes)!"; \
+		exit 1; \
+	fi
+
+## Runs terraform get and terraform init for env
+init:
```
denibertovic: The Makefile should inclue the `require-SOME_ENV_VAR` target if it uses any env vars.

denibertovic: This helps us to not run things accidentally with an undefined env....who knows what might happen.

examples/kops-kube-vpc/Makefile
```diff
@@ -0,0 +1,80 @@
...
+
+## terraform apply
+apply: check-plan-file
+	@terraform apply tf.out
+
+## Prepare the environment with a subdomain and vpc for kops
+prepare-environment: plan apply
+
+## Generate terraform plan using `kops`
+generate-kops-plan:
```
denibertovic: I'd like if we could follow the same naming that we have in the existing Makefile in our repo. For example `make kops-create-cluster`, `make kops-update-cluster` etc. We should strive to be as consistent as possible to when we switch projects we can still rely on "muscle memory".

examples/kops-kube-vpc/Makefile
```diff
@@ -0,0 +1,80 @@
+# Hardcoding value of 3 minutes when we check if the plan file is stale
+STALE_PLAN_FILE := `find "tf.out" -mmin -3 | grep -q tf.out`
+SHELL := /bin/bash
+
+AWS_AVAILABILITY_ZONE := `terraform output -json | jq '.availability_zone.value' | tr -d '"'`
+KOPS_STATE_BUCKET := `terraform output -json | jq '.state_bucket_id.value' | tr -d '"'`
+KOPS_CLUSTER_NAME := `terraform output -json | jq '.subdomain_zone_name.value' | tr -d '"'`
+KOPS_CLUSTER_ZONE := `terraform output -json | jq '.subdomain_zone_id.value' | tr -d '"'`
+KOPS_AWS_VPC := `terraform output -json | jq '.vpc_id.value' | tr -d '"'`
```
denibertovic: I'd really like it if these were explicitly set in the example/env file rather than being semi automatic like this. The reason being that the tf code will live in the terraform directory and the kubernetes code will live in another directory with it's separate Makefile and we'd like to avoid to semi magic implicit variables like this. Does that make sense in the case of this example ?

aniketd: > Does that make sense in the case of this example ?

I guess I went overboard with DRY. Although this example is TF + Kops, I think, now that you mention this (and on Slack too), that coupling the two this tightly could be less helpful than imagined.

examples/kops-with-vpc/README.md
```diff
@@ -0,0 +1,184 @@
+# Example to test the Kops workflow with an existing VPC on AWS
+
+## Summary
```
borsboom: I think this summary could use an explanation of the motivation for having Terraform handle creating the VPC rather than letting `kops` do that itself.  Basically: this makes it easier to fit a cluster into your existing infrastructure (e.g. you might want the same VPC to container other non-kops resources).

examples/kops-with-vpc/README.md
```diff
@@ -0,0 +1,184 @@
...
+...and modifies the tags on existing subnets to make things easier for discovery within the kubernetes cluster
+
+## Environment creation and deployment
+`NOTE`: For usage with `aws-env` have each command below execute in the form `aws-env -p <your-profile> -- <the-command>`.
+
+### Part 1: Terraform
+
+```
+pushd terraform
+source prod/env
```
borsboom: I'd suggest providing a `template/env` instead of `prod/env`, and then have instructions to copy and modify the template (same applies in the kubernetes part).

examples/kops-with-vpc/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+1. Create the VPC
+    - 1 SSH Keypair to use throughout
+    - 1 VPC, and
+    - 3 subnets.
+
+```
+make generate-ssh-keypair
+make import-ssh-keypair
+make init
```
borsboom: The `make init` step fails because the Terraform state S3 bucket referenced in `main.tf`'s backend doesn't exist.

examples/kops-with-vpc/terraform/main.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  backend "s3" {
```
borsboom: I'm not sure that this example should includ S3 remote state, since setting up remote state should be up to the user.  If it _does_ include the remote state, then the key should depend on the environment name, the bucket name should be customizable, there should be a step in the Makefile/instructions that creates it.

examples/kops-with-vpc/kubernetes/Makefile
```diff
@@ -0,0 +1,96 @@
...
+		exit 1; \
+	fi
+
+## Fetch private/public ssh keys from pw store
+fetch-keys: require-ENVIRONMENT require-CLUSTER_NAME
+	@pass clusters/${CLUSTER_NAME}/keys/admin.pem.gpg > ${ENVIRONMENT}/${CLUSTER_NAME}-admin.pem
+	@pass clusters/${CLUSTER_NAME}/keys/admin.pem.pub.gpg > ${ENVIRONMENT}/${CLUSTER_NAME}-admin.pem.pub
+
+## Create s3 bucket for kops state
+create-s3-bucket: require-ENVIRONMENT require-CLUSTER_NAME
```
borsboom: Should the Terraform create this bucket instead of making it a separate step?

examples/kops-with-vpc/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+make plan
+make apply
+```
+
+### Part 2: Kubernetes (Kops)
+
+```
+popd; pushd kubernetes
+source prod/env
```
borsboom: The VPC ID in the `env` file should be updated with the ID of the VPC created by Terraform before sourcing it.

examples/kops-with-vpc/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+2. Edit the cluster configuration to help `kops` use the existing VPC and its subnets.
+```
+make kops-edit-cluster
+```
+...here is a table of variable names between the `terraform` output and `kops` configuration that should match after editing the `kops` configuration.
+
+|Terraform output|Kops configuration|
+|----------------|------------------|
+|`vpc_id`|`networkID`|
```
borsboom: Since the VPC ID and VPC CIDR are passed to `kops create cluster` from the `env` file, these should not need to be changed here.

examples/kops-kube-vpc/Makefile
borsboom: Here are some more concrete thoughts on how to avoid the copying-and-pasting of values:

- Add a top-level `Makefile` that is in charge of coordinating between the Terraform and the Kubernetes

    - Add a target that writes `kubernetes/<ENVIRONMENT>/tf_env` containing things like the VPC_ID, VPC_CIDR, SUBNET_IDS, SUBNET_CIDRS, and anything else that can be gotten from the `terraform/<ENVIRONMENT>/env` or `terraform output`s.  BTW, you can get the value from a Terraform output using `terraform output <OUTPUT_NAME>` rather than using `jq`.

    - Also add a target that copies the SSH key

- Modify `kubernetes/<ENVIRONMENT>/env` to `source "$(dirname "${BASH_SOURCE[0]}")/tf_env"` and then set any values that aren't in the tf-env.  There probably won't be much left besides the `KOPS_MASTER_*`, `NODE_*`, and `KUBE*` variables.

- In `kubernetes/Makefile`, add a target that uses `kops get cluster fpco-prod-v2.fpcomplete.com -o json`, `jq`, and `kops replace -f` to update the subnet IDs and CIDRs from the env.  This would replace manually editing the cluster manifest.

Thoughts?

examples/kops-with-vpc/kubernetes/Makefile
aniketd: I was on the fence about this one too... Reason I kept it this way was to have dependencies separate between the directories. But for this specific example, it does make sense to have it done in Terraform and update the Kops `env` accordingly.

examples/kops-with-vpc/README.md
```diff
@@ -13,18 +15,21 @@ We then use `kops` to create the kubernetes cluster in this existing VPC, where
...
+6. Files and secrets required to configure kubernetes clusters
 7. Security groups
+
 ...and modifies the tags on existing subnets to make things easier for discovery within the kubernetes cluster
 
 ## Environment creation and deployment
-`NOTE`: For usage with `aws-env` have each command below execute in the form `aws-env -p <your-profile> -- <the-command>`.
+`NOTES`:
+  - For usage with `aws-env` have each command below execute in the form `aws-env -p <your-profile> -- <the-command>`.
+  - The word `template` in the snippets below refers to your target environment. As an example, `template/env` is provided under the respective directories for reference. Feel free to make directories named after your target environment like `production`, `staging`, etc. in both `terraform` and `kubernetes` sub-directories to support multiple target environments.
```
borsboom: I think this can be worded a more strongly: users _should_ copy the template to their own environment name and then modify it, rather than use it as it is.

aniketd:
@denibertovic @borsboom I have updated this PR with the latest changes! :champagne:

examples/kops-with-vpc/README.md
```diff
@@ -53,28 +53,63 @@ make plan
...
 
-### Part 2: Kubernetes (Kops)
-
+### Part 2: Copy relevant values over to the Kubernetes' side
 ```
-popd; pushd kubernetes
+popd
+cp -r kubernetes/template kubernetes/$ENVIRONMENT
+make copy-keys                       # Copy generated .pem files
+make copy OUT=region AS=REGION       # Copy terraform output to kube(kops) environment: `$ENVIRONMENT/env`
```
borsboom: Could this whole list of variables be in the Makefile instead of needing to run each command separately?  One way would be for the Makefile to have a new phony target that runs each of these commands.

examples/kops-with-vpc/README.md
```diff
@@ -53,28 +53,63 @@ make plan
...
+```
+source test/env
 ```
 
 1. Create the cluster configuration
 ```
 make kops-create-cluster
 ```
-2. Edit the cluster configuration to help `kops` use the existing VPC and its subnets.
+2. Use `kops get cluster -o json`, `jq` and `kops replace -f` to update the cluster, like so
```
borsboom: Could this command be moved into the Makefile?

aniketd:
Finally! :+1: :heart:

PR #184, opened 2018-12-18
new-module: init-snippet-install-mosh

aniketd:
This is a new module to install `mosh` on boot, just like we have for `aws-cli`.


ketzacoatl:
This isn't the goal of the init modules, mosh would be better installed by CM tooling when/where it is desired.

Issue #188, opened 2019-03-07
new example: demonstrate basic ASG integration with Lifecycle Hooks


ketzacoatl:
Autoscaling Groups can be setup with one or more Lifecycle hooks, which can then be used to take specific actions as EC2 instances pass through the various states of their "lifecycle". In essence, lifecycle hooks put EC2 instances in a "paused/pending" state before going into service, and before terminating. However, implementing the full picture requires linking together several components, and there are some options for how it's all setup.

The goal of this ticket is to produce an initial example that we can use to explore specific use cases and implementation ideas when in those scenarios. Future tickets can add to this example or create additional variations on the example to cover other use cases.

We're already using `user_data` for init (though lifecycle hooks could still bring improvements there), so at least right now, we're primarily interested in taking actions before terminating an EC2 instance.

Specifically, we'd like build out coverage for:
* de-registering EC2 instances from an ELB or ALB's target group before terminating each instance
* when using a cluster management tool (Kubernetes or Nomad, for example), we should use the manager's APIs to mark the instance to terminate as in-eligible for new tasks and should also be drained of currently running tasks.

We are interested in two main ways to run actions on lifecycle events:
* a lambda function
* some script or tool that runs from an EC2 instances

Interacting with the AWS API is a pretty easy thing to do with lambda, but using tools like `kubectl` to interact with cluster management APIs is not. Due to our use of cluster management tools, running the tool/script from an EC2 instance is more useful and desirable than running from lambda. However, we would like to demonstrate how to use lambda for running actions on lifecycle events.

### Requirements

* Must use an existing example which has an ALB and some webapp servers in an ASG.
* Could demo how to run a "cleanup action" on a EC2 webapp instance when it is in the terminated lifecycle hook - that is, each instance would run an action on itself when it is terminating.
* Should add an EC2 instance that can be used to run actions on lifecycle events (let's call it the "Monitor")
* Must give that EC2 instance the permissions it needs (to access resources on AWS) via IAM instance profile
* Should a tool/script/agent (lifecycled or some other demo script), on the Monitor, to handle lifecycle events for all the webapp servers. Do something like de-register it from the ALB
* If and where it makes sense, should encapsulate lifecycle components in modules (or at least note that as a task for later).
* Should leave running actions from lambda as a task for another example, or to include in this example later, unless everything else is complete and there is time available in this issue.

### Resources

[...]

PR #189, opened 2019-03-12
WIP: Demonstrate basic ASG integration with lifecycle hooks

psibi:


ketzacoatl:
As a first pass, getting an example working, this looks great!

Reviewing in detail, I would recommend the following updates:

* Lifecycled should be installed from https://github.com/buildkite/lifecycled/releases
  * Some clients will want to build/publish all artifacts used, while most will want to use what is released from upstream.
* ATM all the lifecycled stuff is in a single module, though that would not be useful to a client project. It would be better to break up specific functionality into separate modules, or leave some of the resources in the root of the example
  * We already have an `asg` module, but there are going to be issues with the `initial_lifecycle_hook` block. TF 0.12.x would address that. For now, we might want to just have two `asg` modules, merging them after we update the codebase to support `0.12.x`
  * Separate the cloud init from the ASG it is applied to, those should be de-coupled, as we will run _different_ init on a client
  * The IAM resources should be separated from the ASG itself
* Using cloudwatch here might be useful for debugging the example, but makes the code less reusable, so I would drop that.
* Overall, an example should provide guidance on what is needed, but also help reduce the amount of copy/paste - eg, grab these modules and you are good. ATM there is too much that we would need to change when applying to a client project, but the updates noted above ought to help clear that up.
* We want the example to primarily focus on what is _required_ to run lifecycled.

It's not necessary now, but just as a heads up, I'll want to squash down the commits (can be more than 1 commit, but should be a clear logical flow without the sausage making)..

It would also be great to add to this PR, some example output or screenshots that confirm the lifecycle hooks are functional.

psibi:
Some updates:
* lifecycled is now installed from github release
* Removed cloudwatch dependency in the infrastructure

But there are still some points which needs to addressed from your above comments which I will work on.

psibi:
Some updates from last changes:

* Remove the lifecycled-module
* Introduce new module named asg-lifecycle for Auto scaling group with lifecycle hooks enabled
* De-register the instance from ELB from the lifecycle handler

ketzacoatl:
This resolves https://github.com/fpco/terraform-aws-foundation/issues/188

PR #206, opened 2019-07-08
modules: Add `init-snippet-gitlab-docker-compose`.

aniketd:
The motivation behind this is the format for:
`GITLAB_OMNIBUS_CONFIG`
This has to be on a single line when using an
rc.local file. Multi-line support is available
in a `docker-compose.yml` file.

- [x] Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Modules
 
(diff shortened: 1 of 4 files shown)


aniketd:
I'd be happy to replace that one with this. :+1:

ketzacoatl:
Closing for now, we'll update the existing module separately.

PR #208, opened 2019-07-12
WIP: Vault and IAM integration

psibi:
---
name: Pull request template
---

- [ ] Update the changelog


modules/vault-iam/Makefile
```diff
@@ -0,0 +1,43 @@
...
+	@echo "Please use \`make <target>' where <target> is one of\n\n"
+	@awk '/^[a-zA-Z\-\_0-9]+:/ { \
+		helpMessage = match(lastLine, /^## (.*)/); \
+		if (helpMessage) { \
+			helpCommand = substr($$1, 0, index($$1, ":")-1); \
+			helpMessage = substr(lastLine, RSTART + 3, RLENGTH); \
+			printf "%-30s %s\n", helpCommand, helpMessage; \
+		} \
+	} \
+	{ lastLine = $$0 }' $(MAKEFILE_LIST)
```
ketzacoatl: We would want to drop the Makefile from here in the module. The Makefiles are helpful in the examples but not here. Alternatively, if you are using a Makefile here for R&D, we should add a generic makefile at the top of the repo that can be used with any module, and in the way you use (or a similar generic solution).

modules/vault-iam/main.tf
```diff
@@ -0,0 +1,44 @@
+resource "aws_iam_user" "vault_user" {
+  name = "vault_user"
+  tags = {
+    "user" = "vault"
+  }
+}
+
+resource "aws_iam_access_key" "vaultkey" {
+  user = "${aws_iam_user.vault_user.name}"
+}
```
ketzacoatl: I'm generally against using Terraform to pass around these credentials. I could be convinced to use `count` and make this optional.

modules/vault-iam/main.tf
```diff
@@ -0,0 +1,44 @@
...
+
+resource "aws_iam_user_policy" "vault_user_policy" {
```
ketzacoatl: We could reuse the general idea here, so I would recommend we consider renaming the module be a generic module, which we can then use in your vault example with the `vault` namespace. Alternatively, if this module needs to have a custom list of actions, we could create a generic module that provides "admin on IAM within this namespace", where the namespace is provided as a variable, and separate from this `vault-iam` module.

modules/vault-iam/main.tf
psibi: Fair enough. I will make this as optional.

modules/vault-iam/main.tf
psibi: I have extracted a generic module named `iam-user-policy` out of this. I have moved the `vault-iam` module under `examples` where it uses this module to create an user and an associated policy.

modules/vault-iam/main.tf
psibi: I make this optional now with the default behaviour not creating any keys.

modules/vault-iam/Makefile
psibi: This module itself has been removed (and hence the Makefile also). More details in this comment: https://github.com/fpco/terraform-aws-foundation/pull/208#discussion_r304760849

ketzacoatl:
I'd like to make the following changes before merging:
* make `vault-iam` a module
* use a data source for the policy JSON
* parametize the `vtest` in `arn:aws:iam::${data.aws_caller_identity.current.account_id}:user/vtest-*` with a variable that defaults to `vault`
* update the `vault-s3-private` example to use the `vault-iam` module
* drop the `aws` provider from the `vault-aws-backend` module
* consider dropping the `vault_aws_secret_backend` resource and use a make target (this resource would include admin secrets in Terraform's state file)

psibi:
@ketzacoatl On further thought, I think we should remove the entire `vault-aws-backend` module itself from terraform (because we don't want admin secret in Terraform's state file and just having `vault_aws_secret_backend_role` resource isn't useful). 

Also, I don't think it is useful for us to make `vault-s3-private` use `vault-iam` module, because they are not dependent. Only the running vault server's secret engine needs the access keys from the `vault-iam` module. No other output from `vault-iam` is actually needed for it in `vault-s3-private`. I think the right way for us to proceed would be:

* Have make target step for the `vault-iam` module. This step will create access keys which has to exposed via environment variables by us manully for the next step.
* Have another make target step (like what you have suggested) to create AWS secret engine path and create an appropriate role for this. (This step will completely replace the `vault-aws-backend` module). This step will require proper environment variable for access keys to be present for it to work.

What do you think ?

ketzacoatl:
Yes, that sounds good.

I would comment though, that if there are no secrets, and we can do it in Terraform, we ought to. Some of the vault configuration can be done in Terraform and doesn't include secrets. Those should stay in Terraform. IDK if that is relevant to the `vault-aws-backend` module, I haven't reviewed the code in a few days.

ketzacoatl:
@psibi are these updates WIP or is it ok to have one of our engineers make those updates?

PR #213, opened 2019-07-17
New  modules for IAM users/groups

ketzacoatl:

- [X] Update the changelog
- [X] Make sure that the module is added to [tests/main.tf](https://github.com/fpco/terraform-aws-foundation/blob/master/tests/main.tf)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Modules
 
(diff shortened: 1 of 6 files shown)


ketzacoatl:
ATM linting fails on CI on this module b/c CI is using Terraform 0.11.x and this module is written for 0.12.x

PR #219, opened 2019-07-18
Initial support for Terraform v0.12.x

ketzacoatl:
Thanks for @Magicloud for tackling the repo-wide upgrade! The MR has been testing well, but it's likely imperfect. I'd rather release and fix and release updates, than hold this back. 

- [X] Update the changelog

### DIFF
--- a/examples/gitlab-ha/main.tf
+++ b/examples/gitlab-ha/main.tf
@@ -21,13 +21,14 @@
  */
 
 provider "aws" {
-  region = "${var.region}"
+  region = var.region
 }
 
-data "aws_availability_zones" "available" {}
+data "aws_availability_zones" "available" {
+}
 
 data "aws_route53_zone" "zone" {
-  name = "${var.dns_zone_name}."
+  name         = "${var.dns_zone_name}."
   private_zone = false
 }
 
@@ -37,18 +38,19 @@ module "ubuntu-xenial-ami" {
 }
 
 resource "aws_acm_certificate" "cert" {
-  domain_name = "*.${var.dns_zone_name}"
+  domain_name       = "*.${var.dns_zone_name}"
   validation_method = "DNS"
 }
+
 resource "aws_key_pair" "main" {
-  key_name   = "${var.name}"
-  public_key = "${file(var.ssh_pubkey)}"
+  key_name   = var.name
+  public_key = file(var.ssh_pubkey)
 }
 
 # S3 bucket for the Docker Registry (running in gitlab) to store Docker Images
 module "docker-registry-s3-storage" {
   source      = "../../modules/s3-remote-state"
-  bucket_name = "${var.registry_bucket_name}"
+  bucket_name = var.registry_bucket_name
   versioning  = "false"
   principals  = []
 }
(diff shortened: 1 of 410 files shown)


Magicloud:
I will make the changelog.

For lint issue, where did you get it? https://gitlab.fpcomplete.com/fpco-mirrors/terraform-aws-foundation/-/jobs/12750 is the 0.12 branch built with new default-build-image. Lint seems passed.

Magicloud:
One thing I am not sure is the hardcoded version of terraform in Makefile. Why 0.12.2 when latest release is 0.12.4? I think this also affects default-build-image. If we want to stick to a non-latest version, I'd suggest to have a KV DB to store those kind of "company preference". So every where, it is the right version.

Magicloud:
The review system is bit of weird, so I post some comments without a link to source.

In _examples/load-asg/main.tf_, around line 105, you removed the TODO generated by TF. That is fine. But seems like you also removed two other comments, started with "Ensure", and "ELBs". Just to confirm it is on purpose.

Magicloud:
And there are multiple syntax error, for example, _examples/kube-stack-public/kube.tf_, line 24.

It is a lint for one element list. But the closing bracket is gone.

ketzacoatl:
Is the code review physically not working? It would be helpful if you could put the comments at the relevant code in the future. Thank you for the review, and you are correct, removing the TODO should not have removed the next line.

Magicloud:
Sorry, get it working now. Posted my comments.

And for syntax error, I think I made mistake. I said multiple, but reviewing today gets only one.

modules/single-port-sg/main.tf
```diff
@@ -46,24 +46,24 @@ locals {
...
-  count             = "${local.tcp}"
+  count             = local.tcp
   type              = "ingress"
   description       = "${var.description} (tcp)"
-  from_port         = "${var.port}"
-  to_port           = "${var.port}"
+  from_port         = var.port
+  to_port           = var.port
   protocol          = "tcp"
   cidr_blocks       = ["${var.cidr_blocks}"]
```
qrilka: this one should change as well

ketzacoatl: Yes, definitely. Good catch, I have updated that.

ketzacoatl:
@qrilka / @Magicloud ok, I have updated the branch again. Please review one last time, give approval if that is ready, and we can merge.

examples/gitlab-ha/main.tf
```diff
@@ -21,13 +21,14 @@
...
+  region = var.region
 }
 
-data "aws_availability_zones" "available" {}
+data "aws_availability_zones" "available" {
+}
 
 data "aws_route53_zone" "zone" {
-  name = "${var.dns_zone_name}."
+  name         = "${var.dns_zone_name}."
```
psibi: The trailing dot seems to be accidental here.

modules/bind-server/main.tf
```diff
@@ -27,73 +27,73 @@ resource "aws_instance" "bind" {
...
 
-  tags {
-    Name = "${length(var.names) == 0 ? format("%s-%02d", var.name, count.index + 1) : var.names[count.index]}"
+  tags = {
+    Name = length(var.names) == 0 ? format("%s-%02d", var.name, count.index + 1) : var.names[count.index]
   }
 
   lifecycle {
-    ignore_changes = ["key_name"]
+    ignore_changes = [key_name]
```
psibi: Is this right or should it be `var.key_name` ?

examples/gitlab-ha/main.tf
ketzacoatl: I believe it was added assuming that the operator would provide `foo.bar.tld`, but we'd need to query with `foo.bar.tld.` - this is a DNS thing

modules/bind-server/main.tf
ketzacoatl: I think it should be `"key_name"`, good catch.

PR #221, opened 2019-07-19
New module ami-centos. Fixes #218.

### LINKED ISSUE #218: new module: ami-centos
Similar to the existing `ami-ubuntu` module, we could use a `ami-centos` variant.

## Requirements

* Review source accounts, choose an upstream carefully (the Ubuntu variant uses Canonical).
* Parametize with focus on the OS release/version, keeping it simple.

Magicloud:
Add a new module to lookup the CentOS AMI we want.

- [x] Update the changelog
- [x] Make sure that there is an up to date example for your code:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,10 @@
 
 * `single-port-sg`: Add support for UDP rules
 * `kube-controller-sg`: Update to sync with `single-port-sg` module
+* `ami-centos`: simple module to get the ami id of the specified release of CentOS.
 
 ### Examples
-
+* `ami-centos-test`: New testing example for `ami-centos`.
 
 # v0.8.3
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,9 +6,10 @@
 
 * `single-port-sg`: Add support for UDP rules
 * `kube-controller-sg`: Update to sync with `single-port-sg` module
+* `ami-centos`: simple module to get the ami id of the specified release of CentOS.
 
 ### Examples
-
+* `ami-centos-test`: New testing example for `ami-centos`.
```
ketzacoatl: Should we just the module in the `vpc-scenario-2` example?

Magicloud: I think that depends on the purpose. `vpc-scenario-2` sure can work with this module. But as a test, it involves other modules, not quite a "unit" test.

ketzacoatl: I'd agree with you on that. The one issue I see is that we would need to put more time into maintaining each of the unit tests than we are now.

PR #223, opened 2019-07-20
New module: autoscaling-policy-metric-alarm-pair

### LINKED ISSUE #195: new module: autoscaling-policy-metric-alarm-pair
#156 created a new example that demonstates how to use AWS autoscaling policies and metric alarms to scale an ASG up and down based on the average CPU and memory load across the cluster. That example is great, but did not get into making a module out of the key resources for dynamic, load-based, auto scaling.

## Requirements

The purpose of this ticket is to create a module that would be re-used several times in that example. EG, instead of the following:

```
resource "aws_autoscaling_policy" "mem_up" {
  name = "${var.name}-asg-policy-mem-up"
  autoscaling_group_name = "${module.web-asg.name}"
  adjustment_type     = "ChangeInCapacity"
  cooldown            = 300
  scaling_adjustment  = 3
  policy_type         = "SimpleScaling"
}

# Cloudwatch Monitor Memory Up
resource "aws_cloudwatch_metric_alarm" "web_mem_scale_up" {
  alarm_name = "${var.name}-mem-scale-up-alarm"
  comparison_operator = "GreaterThanOrEqualToThreshold"
  evaluation_periods = "3"
  metric_name = "MemoryUtilization"
  namespace = "AWS/EC2"
  period = "60"
  statistic = "Average"
  threshold = "60"
  dimensions {
    AutoScalingGroupName = "${module.web-asg.name}"
  }
  alarm_actions = ["${aws_autoscaling_policy.mem_up.arn}"]
}
```

We'd have something like:

```
module "auto-scale-mem-up" {
  source = "../../modules/autoscaling-policy-metric-alarm-pair"

  asg_name     = "${module.web-asg.name}"
  metric_name = "MemoryUtilization"
}
```

The module should parametize the details of those two resources by defining a bunch of variables w/ defaults. Require only the `asg_name` and `metric_name` parameters.

Lastly, update the example to use the new module. There should be 4 occurances of the module, one for each up/down on the cpu/memory metrics.

### LINKED ISSUE #222: render-ssh-config could not run for examples/load-asg
In server-ip.sh, it filters the instance by "tag:Name". But the "server_name" output from `terraform` is actually "tag:aws:autoscaling:groupName". Hence the `describe-instances` returns nothing.

Magicloud:
Tested following the Readme of load-asg.

No worries on the terraform version. Will update when v0.12.x is merged. Fixes #195 . Fixes #222 .

### DIFF
--- a/modules/autoscaling-policy-metric-alarm-pair/main.tf
+++ b/modules/autoscaling-policy-metric-alarm-pair/main.tf
@@ -1,0 +1,109 @@
+variable "name" {
+  description = "Prefix of the AutoScaling Policy."
+}
+
+variable "asg_name" {
+  description = "Name of the AutoScaling Group that this policy would apply to."
+}
+
+variable "metric" {
+  description = "CPUUtilization or MemoryUtilization"
+}
+
+variable "up_evaluation_periods" {
+  default = "3"
+  description = ""
+}
+
+variable "up_period" {
+  default = "60"
+  description = ""
+}
+
+variable "up_threshold" {
+  default = "60"
+  description = ""
+}
+
+variable "up_scaling_adjustment" {
+  default = "3"
+  description = ""
+}
+
+variable "up_cooldown" {
+  default = "300"
+  description = ""
+}
+
+variable "down_evaluation_periods" {
+  default = "3"
+  description = ""
+}
+
+variable "down_period" {
+  default = "60"
+  description = ""
+}
+
+variable "down_threshold" {
+  default = "30"
+  description = ""
+}
+
+variable "down_scaling_adjustment" {
+  default = "2"
+  description = ""
+}
+
+variable "down_cooldown" {
+  default = "300"
+  description = ""
+}
+
+resource "aws_autoscaling_policy" "scale_up" {
+  name                   = "${var.name}-asp-up"
+  autoscaling_group_name = "${var.asg_name}"
+  adjustment_type        = "ChangeInCapacity"
+  cooldown               = "${var.up_cooldown}"
+  scaling_adjustment     = "${var.up_scaling_adjustment}"
+  policy_type            = "SimpleScaling"
+}
+
+resource "aws_cloudwatch_metric_alarm" "scale_up" {
+  alarm_name          = "${var.name}-scale-up-alarm"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "${var.up_evaluation_periods}"
+  metric_name         = "${var.metric}"
+  namespace           = "AWS/EC2"
+  period              = "${var.up_period}"
+  statistic           = "Average"
+  threshold           = "${var.up_threshold}"
+  dimensions {
+    AutoScalingGroupName = "${var.asg_name}"
+  }
+  alarm_actions = ["${aws_autoscaling_policy.scale_up.arn}"]
+}
+
+resource "aws_autoscaling_policy" "scale_down" {
+  name                   = "${var.name}-asp-down"
+  autoscaling_group_name = "${var.asg_name}"
+  adjustment_type        = "ChangeInCapacity"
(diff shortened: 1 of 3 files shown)


modules/autoscaling-policy-metric-alarm-pair/main.tf
```diff
@@ -0,0 +1,59 @@
...
+  statistic           = "Average"
+  threshold           = "60"
+  dimensions {
+    AutoScalingGroupName = "${var.asg_name}"
+  }
+  alarm_actions = ["${aws_autoscaling_policy.scale_up.arn}"]
+}
+
+resource "aws_autoscaling_policy" "scale_down" {
+  name                   = "${var.name}-asp-down"
```
ketzacoatl: `asg` in this and the `scale_up`

modules/autoscaling-policy-metric-alarm-pair/main.tf
```diff
@@ -0,0 +1,59 @@
...
+  alarm_name          = "${var.name}-scale-down-alarm"
+  comparison_operator = "LessThanOrEqualToThreshold"
+  evaluation_periods  = "3"
+  metric_name         = "${var.metric}"
+  namespace           = "AWS/EC2"
+  period              = "60"
+  statistic           = "Average"
+  threshold           = "30"
+  dimensions {
+    AutoScalingGrodownName = "${var.asg_name}"
```
ketzacoatl: Is this parameter correct or misspelled?

modules/autoscaling-policy-metric-alarm-pair/main.tf
```diff
@@ -0,0 +1,59 @@
...
+
+resource "aws_cloudwatch_metric_alarm" "scale_down" {
+  alarm_name          = "${var.name}-scale-down-alarm"
+  comparison_operator = "LessThanOrEqualToThreshold"
+  evaluation_periods  = "3"
+  metric_name         = "${var.metric}"
+  namespace           = "AWS/EC2"
+  period              = "60"
+  statistic           = "Average"
+  threshold           = "30"
```
ketzacoatl: I believe it would be good for us to parametize these details we would likely want to tweak. Eg the `evaluation_periods`, `period`, `threshold`, `scaling_adjustment` , `cooldown`, should all be variables with reasonable defaults.

Also, the module creates scale up and down resources in the module, but in the example env, the module is used twice, once for scaling up and once for scaling down - We should pick one.. either the module covers both scenarios or is used twice to cover each scenario. I believe we would be slightly better to have the module focus on one scenario at a time, and used multiple times if that is needed.

modules/autoscaling-policy-metric-alarm-pair/main.tf
Magicloud: Oh, I did not mean "asg", I mean "AutoScaling Policy".

modules/autoscaling-policy-metric-alarm-pair/main.tf
Magicloud: OK.

modules/autoscaling-policy-metric-alarm-pair/main.tf
Magicloud: Interesting, running this did not get me the error.

modules/autoscaling-policy-metric-alarm-pair/main.tf
ketzacoatl: It's very likely that you would want slightly different parameters for scaling up / down. Eg, maybe there is a shorter cooldown period, or usually there is a longer period for the metrics evaluation. At the same time, you are right, you would use both the up and down together, but I am just recommending the module be simpler and used/called more times - eg instead of one module with both scale up and down, there is one general module we call twice when we want to use it.

Magicloud:
Close in favor of #228 which is for 0.12.

PR #224, opened 2019-07-20
tests: disable v0.12.x modules until we merge v0.12 branch

ketzacoatl:
There are 2 modules that are for v0.12.x only, but they were merged to master a little early. That's ok, but they are making the tf linting fail (as that is focusing on v0.11 syntax). These will be re-enabled after the v0.12.x branch is merged.


ketzacoatl:
CI passed!

PR #231, opened 2019-07-24
New: module iam-instance-profile

### LINKED ISSUE #48: light refactor for EC2 instance profiles and IAM policies
For modules that use IAM profiles and are generally hooked up to EC2 instance profiles, our modules should:

* Assume the user would create an EC2 instance profile, a role for the instance, and some number of policy attachments.
* Provide the components to make that easy. For example, the `persistent-ebs` module create the profile for you and gives it an inconsistent name - modules should create resources with names and under the assumption the instance profile may get several different policies giving it permissions to different resources on AWS.
* export the right outputs so as to facilitate that assumption above

Magicloud:
Fixes #48.

The module was tested in EIP project, although that did not work through. In the generated instance, I could run `aws` to attach EBS, and associate EIP, which proved that the permissions were given correctly. And outside EIP project, running the single-node-asg gave instance without associate-address permission, which proved that no wrong permission was given.

- [X] Update the changelog
- [x] Make sure that there is an up to date example for your code:


ketzacoatl:
I have WIP updates to this PR. I will push those ASAP.

PR #233, opened 2019-07-31
Bugfix: Something missed in the major 0.12 upgrade.

Magicloud:
In the VPC Scenario 2 example, we have some lingering updates to make from the v0.12.x upgrade:

* Lifecycle does not exist now.
* Drop extra `[]`
* Fixup 0.11.x syntax in `bastion.tf`
* Update the `Makefile` in the process of testing


ketzacoatl:
> Lifecycle does not exist now.

is there a replacement for `ignore_changes`?

Magicloud:
> > Lifecycle does not exist now.
> 
> is there a replacement for `ignore_changes`?

I am not sure. Let me check.

PR #234, opened 2019-07-31
New: RDS Aurora cluster

Magicloud:
Creating RDS cluster via AWS webui is pretty simple, but the model underneath is not. So make this module to indicate what is necessary for bringing up an Aurora cluster.

Example rds-aurora-test shows how to use this module.

---
name: Pull request template
---

- [ ] Update the changelog

### DIFF
--- a/modules/rds-aurora/main.tf
+++ b/modules/rds-aurora/main.tf
@@ -1,0 +1,85 @@
+variable "azs" {
+  type        = list(string)
+  description = "Availibility zones for cluster to use."
+}
+
+variable "db_name" {
+  type        = string
+  description = "Database name."
+}
+
+variable "db_password" {
+  type        = string
+  description = "Database master password."
+}
+
+variable "db_username" {
+  type        = string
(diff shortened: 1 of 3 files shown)


modules/rds-aurora/main.tf
```diff
@@ -0,0 +1,85 @@
...
+}
+
+variable "sg_ids" {
+	type = list(string)
+	description = "Security group ids that controls the access to database."
+}
+
+variable "subnet_ids" {
+	type = list(string)
+	description = "Subnet ids for database to live in. Better be private subnets."
```
qrilka: As I remember it from setting up RDS a couple of weeks ago this list is supposed to contain at least 2 elements even for single-az deployment - https://forums.aws.amazon.com/thread.jspa?threadID=151938
What do you think about adding this link as a reminder?

Magicloud: Yes, that is true. At least two subnets are required unless there is only one AZ in the region. I will put the link.

PR #237, opened 2019-08-04
Bugfix and improvement.

Magicloud:
- ASG user_data cannot be empty string. So rework default value.

- ASG does not take target group for ALB usage.

---
name: Pull request template
---

- [ ] Update the changelog

### DIFF
--- a/modules/asg/variables.tf
+++ b/modules/asg/variables.tf
@@ -38,7 +38,7 @@ variable "placement_group" {
 }
 
 variable "user_data" {
-  default     = ""
+  default     = "#!/bin/sh"
   description = "The user_data string to pass to cloud-init"
   type        = string
 }
--- a/modules/asg/variables.tf
+++ b/modules/asg/variables.tf
@@ -77,7 +77,7 @@ variable "elb_names" {
 
 variable "alb_target_group_arns" {
   default     = []
-  description = "list of target_group for application load balancer to associate with the ASG (by ARN)"
+  description = "A list of aws_alb_target_group ARNs, for use with Application or Network Load Balancing."
   type        = list(string)
 }
 


Magicloud:
@qrilka I knew you used web-alb module. So please give advise here.

qrilka:
Nah, in the end it didn't work out because of requirement of preexisting ACM certificate. So in the end I stole some code from https://gitlab.fpcomplete.com/greatcall/consumer-env/tree/master/terraform/modules/env-wrapper
I'll take a look into this tomorrow

examples/alb-test/tester.tf
```diff
@@ -0,0 +1,64 @@
...
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+}
+
+module "alb" {
+  source        = "../../modules/alb"
+  ports         = ["80"]
+  vpc_id        = module.vpc.vpc_id
+  sg_ids        = [aws_security_group.wide_open.id]
+  identifier    = "albest"
```
qrilka: does this identifier have any special meaning? I guess the first part is ALB but what does suffix mean? Shouldn't it be "alb_test" as above?

examples/alb-test/tester.tf
```diff
@@ -0,0 +1,64 @@
...
+  subnet_ids    = module.vpc.public_subnet_ids
+  ports_secured = []
+  dns_name      = "shida.dev-sandbox.fpcomplete.com"
```
qrilka: @Magicloud do we need to hardcode your name here and for the key name above?

examples/alb-test/tester.tf
```diff
@@ -0,0 +1,64 @@
...
+  acm_arn       = ""
+}
+
+data "aws_route53_zone" "dev-sandbox" {
+  name = "dev-sandbox.fpcomplete.com."
+}
+
+resource "aws_route53_record" "shida" {
+  zone_id = data.aws_route53_zone.dev-sandbox.zone_id
+  name    = "shida.${data.aws_route53_zone.dev-sandbox.name}"
```
qrilka: same here

examples/alb-test/tester.tf
Magicloud: Ye.... There is a 6 char limit....

examples/alb-test/tester.tf
Magicloud: It does not. I'd like this to be a unit test and planable without human interfere, so the code can be more robust.. But that would require a framework, that provides necessary variables in CI. Nothing done on that yet.

examples/alb-test/tester.tf
qrilka: Maybe "albtst"? Sounds a bit more readable, BTW where does the limit come from?

examples/alb-test/tester.tf
qrilka: If we don't have a good way resolve this then probably my comment could be ignored...

ketzacoatl:
When using ALB's elsewhere, I generally found it best to keep the ALB separate from the Target Groups. Target Groups are more closely associated with ASGs (1:1). It's also questionable to combine the listeners with ALB.

examples/alb-test/tester.tf
Magicloud: OK. For the limitation, I guess it is something missed in TF doc. For example, target group, says clearly that name_prefix has 6 char limitation. But for lb, it says nothing, but when planning, it throws error.

Magicloud:
@ketzacoatl I have update the PR for some minor changes underneath. Please review.

ketzacoatl:
@Magicloud can you confirm using this parameter, or using the ELB parameter, don't conflict with each other?

Magicloud:
The PR seems overlap some new code in master. Closing.

PR #238, opened 2019-08-04
minor fixups to single-node-asg

ketzacoatl:


ketzacoatl:
This was merged to master separately (I think by mistake, I ought to have done it through this pr).

Issue #242, opened 2019-08-23
availability zones of asg module should be removed.


Magicloud:
After upgrading to terraform 0.12, it happened to me that code using asg module cannot be applied sometimes, with "inconsistent plan" error. After trying the fix of following issue, it seemed working fine.

https://github.com/terraform-providers/terraform-provider-aws/issues/9622

Magicloud:
@ketzacoatl Please review the reference issue and the purposal before I submit PR.

ketzacoatl:
Yep, that LGTM. Maybe you can have @JoseD92 make the updates?

JoseD92:
@Magicloud I was able to reproduce the "inconsistent plan" error, and deleting the availability zones from the asg module does solve the issue, but as stated in the docs for aws_autoscaling_group, availability zones is Required only for EC2-Classic, the asg module would become useless if working with EC2-Classic, I am not sure if that is a use case for terraform-aws-foundation or not, but in the case it is, I think a better solution would be add warnings in the readme of asg module and add assertions (preferably if the can run on both planning and applying) so that terraform exits with an error and an informative message if availability_zones and vpc_zone_identifier are used at the same time

ketzacoatl:
Great review and comment @joseD92! I think it's ok for us to skip out on support for EC2 classic at this point. We should only see non-classic EC2 from now on.

JoseD92:
I add a MR with the fix

PR #247, opened 2019-09-06
removed availability_zones from asg

JoseD92:
This PR removes availability_zones from the asg module, as https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html says availability_zones and vpc_zone_identifier should not be used together causing issue #242, it is also important to note that availability_zones is required to use EC2-Classic, so this update will make terraform-aws-foundation possibly not work on EC2-Classic

### DIFF
--- a/examples/asg-lifecycle-hooks/main.tf
+++ b/examples/asg-lifecycle-hooks/main.tf
@@ -147,7 +147,6 @@ resource "aws_sns_topic" "main" {
 module "asg" {
   source             	  = "../../modules/asg"
   name_prefix        	  = var.name_prefix
-  azs                	  = local.azs
   elb_names          	  = [aws_elb.web.name]
   subnet_ids         	  = module.vpc.public_subnet_ids
   min_nodes          	  = 1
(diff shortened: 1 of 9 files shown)


JoseD92:
running examples with master branch will result, when applying, in:

```
...
aws_security_group.sg: Creation complete after 5s [id=sg-044dff8dea12d391c]
module.asg.aws_launch_configuration.cluster: Creating...
module.asg.aws_launch_configuration.cluster: Creation complete after 3s [id=terraform-20191112203922703500000001]

Error: Provider produced inconsistent final plan

When expanding the plan for module.asg.aws_autoscaling_group.cluster to
include new values learned so far during apply, provider "aws" produced an
invalid new value for .availability_zones: was known, but now unknown.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

Using the change in this PR allows examples to apply correctly taking with the asg's taking the AZ of the subnets used in creating them.

Magicloud:
Looks good to me. Did not know the PR is still open and I got the issue again.

qrilka:
This broke the tests:
```
qrilka@qdesktop ~/ws/fpco/terraform-aws-foundation/tests $ make plan

Error: Unsupported argument

  on ../modules/consul-leaders/main.tf line 29, in module "leader-asg":
  29:   azs                = ["${var.region}a", "${var.region}c"]

An argument named "azs" is not expected here.

make: *** [Makefile:25: plan] Ошибка 1
```

Issue #248, opened 2019-09-13
Changelog seem not to be up-to-date


qrilka:
Release 0.9.4 is out already but the latest released version in CHANGELOG.md is 0.9.3

ketzacoatl:
@JoseD92 maybe you would be able to update that doc to sync?

mcgirr:
I've added a small update to the changelog to add a version `0.9.6` but we still have a good amount of changes that are still in the unreleased section of the changelog that we should sort out. 

It would be good to figure out which of those belong in which release and add entries for `0.9.4` and `0.9.5` as well since those seem to have been skipped when those versions were released.

qrilka:
1d4949ba2c207d0fd209c29b5d009ef9a349e1fe seems to be a fix for this one

PR #249, opened 2019-09-13
Asg additional block devices

qrilka:
---
name: Pull request template
---

- [ ] Update the changelog

### DIFF
--- a/modules/asg/main.tf
+++ b/modules/asg/main.tf
@@ -94,6 +94,15 @@ resource "aws_launch_configuration" "cluster" {
     volume_size = var.root_volume_size
   }
 
+  dynamic "ebs_block_device" {
+    for_each = var.additional_block_devices
+    content {
+      device_name = ebs_block_device.value.device_name
+      volume_type = ebs_block_device.value.volume_type
+      volume_size = ebs_block_device.value.volume_size
+    }
+  }
+
   lifecycle {
     create_before_destroy = true
   }
--- a/modules/asg/variables.tf
+++ b/modules/asg/variables.tf
@@ -93,6 +93,12 @@ variable "root_volume_size" {
   type        = string
 }
 
+variable "additional_block_devices" {
+  default     = []
+  description = "Additional EBS volumes to use as instance block devices"
+  type        = list(object({device_name = string, volume_type = string, volume_size = string }))
+}
+
 // List of maps, as extra tags to append to the Auto-Scaling Group
 variable "extra_tags" {
   default     = []


ketzacoatl:
This change LGTM. Let's get it tested, merged, and released.

qrilka:
I see additional block device available in out current project, do we need some additional testing here?
I wanted to add a line into the changelog but I guess it makes sense to address #248 first

qrilka:
@ketzacoatl we use this module for quite some time already and it seems to work fine. Are there any additional steps required to get this merged?

ketzacoatl:
@qrilka apologies for overlooking the previous question. Can you paste/screenshot/etc a working example/etc that demonstrates the update is functional? That is what we're looking for in general, for all PRs (demonstrate the update works in common use cases, and doesn't break existing use by surprise).

ketzacoatl:
> Should I create a separate ASG in dev-sandbox?

Deploying an ASG to dev-sandbox is fine to do, sure.

Deploying the gitlab example ought to be a good test as well, that should be using all the pieces here and give you an easy way to confirm correctness/functioning. Using an ASG of your own is also ok if you prefer.

>  Will a plan be enough?

A plan is a good start, though that's not as ideal as a full test.

I'd even argue it's worth hooking up the gitlab or another couple examples to our Gitlab for CI and make the test easier to run a lot. We started down this path for foundation in the past (with some S3 tests), but we stopped short. I have also set this up for past projects (the legacy vpc example was one), and it worked sooooo great, I loved having that feedback. If we want to hookup one or more of the examples to Gitlab, we should scope that out in a new ticket.

qrilka:
@ketzacoatl Gitlab examples seem to be not sufficient here as they don't use `asg` directly and also they don't use this new feature. So I went with a custom Terraform file - https://gist.github.com/qrilka/4e1f674ee28bd348e24fc3bd10318688

The resulting instances with (or without) extra EBS could be seen at https://imgur.com/a/4Q064jt

psibi:
@qrilka / @ketzacoatl While we add volumes to it, this doesn't attach the volumes to the instances in the ASG. Also, it seems (for valid reasons) we can attach an ebs volume to only one instance at a time. Do you think we should handle this also ?

qrilka:
@psibi EBS volume specified in a launch configuration creates a new one and attaches it automatically. See e.g. "Additional EBS block devices to attach to the instance. " for `ebs_block_device` at https://www.terraform.io/docs/providers/aws/r/launch_configuration.html
With a launch configuration you attach a volume at launch time. I guess in your question with "attaching" you mean doing that at runtime (even during init)?

ketzacoatl:
w/ Additional EBS Volumes:

image

No Additional Volumes:

image

PR #251, opened 2019-09-17
modules/stp-m-infra: Add iam:Get{User,Role,Policy} for power-users

aniketd:
Background:
The power-user role may have access to change, modify all parts of the
infrastructure except only all IAM related configuration. In certain
situations, like having to modify S3 buckets created by it's own user
the "iam:Get{User,Role,Policy}" permissions become necessary.

This commit grants only those three permission to the power-user role,
or, members of that group.

-----
- [X] Update the changelog
- [X] Make sure that the module is added to [tests/main.tf](https://github.com/fpco/terraform-aws-foundation/blob/master/tests/main.tf)
- [X] Make sure that there is an up to date example for your code:
- [X] Make sure that there is a manual CI trigger that can test the deployment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Modules
 
+* `setup-meta-infrastructure`: add "iam:Get{User,Role,Policy}" permissions for power-users
 * `iam-instance-profile`: Add role ouput for IAM instance profile module.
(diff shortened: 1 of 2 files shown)


ketzacoatl:
I addressed the conflict with the changelog update and rebased to merge and close.

PR #252, opened 2019-09-18
Add multip-port-sg module

qrilka:
---
name: Multip-port security group module
---

- [ ] Update the changelog

### DIFF
--- a/modules/multi-port-sg/README.md
+++ b/modules/multi-port-sg/README.md
@@ -1,0 +1,3 @@
+## Multiple Port Security Group Rule
+
+Create an `aws_security_group_rule` to allow ingress on some ports.
--- a/modules/multi-port-sg/main.tf
+++ b/modules/multi-port-sg/main.tf
@@ -1,0 +1,57 @@
+/**
+ * ## Multiple Ports Security Group Rule
+ *
+ * Create an `aws_security_group_rule` to allow ingress on some ports.
+ *
+ */
+
+variable "security_group_id" {
+  description = "security group to attach the ingress rules to"
+  type        = string
+}
+
+variable "cidr_blocks" {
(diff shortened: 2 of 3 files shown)


qrilka:
The main motivation for this module is the fact that TF 0.12 doesn't allow for_each to be used with modules (though the argument is already reserved for future versions). Because of that `single-port-sg` can't be used with a list of ports to open and this module allows this to be done.

modules/multi-port-sg/README.md
```diff
@@ -0,0 +1,3 @@
+## Multiple Port Security Group Rule
+
+Create an `aws_security_group_rule` to allow ingress on some ports.
```
ketzacoatl: Please add an example of how you intend the module to be used.

modules/multi-port-sg/main.tf
```diff
@@ -0,0 +1,57 @@
...
+  type        = set(string)
+  default     = []
+}
+
+variable "udp_ports" {
+  description = "UDP ports to open"
+  type        = set(string)
+  default     = []
+}
+
```
ketzacoatl: Separate the above into a `variables.tf`

modules/multi-port-sg/main.tf
```diff
@@ -0,0 +1,57 @@
...
+resource "aws_security_group_rule" "udp-ingress" {
+  for_each          = var.udp_ports
+  type              = "ingress"
+  description       = "${var.description} (udp)"
+  from_port         = each.value
+  to_port           = each.value
+  protocol          = "udp"
+  cidr_blocks       = var.cidr_blocks
+  security_group_id = var.security_group_id
+}
```
ketzacoatl: This looks great!

PR #256, opened 2019-10-02
nat-gateways: use non-inline route

borsboom:
- [x] Update the changelog
- N/A Make sure that modules and files are documented. This can be done inside the module and files.
- N/A Make sure that new modules directories contain a basic README.md file.
- N/A Make sure that the module is added to [tests/main.tf](https://github.com/fpco/terraform-aws-foundation/blob/master/tests/main.tf)
- N/A Make sure that there is an up to date example for your code:


lpaulmp:
Looks good and passed the test

PR #262, opened 2019-10-04
Format all terraform sources.

Magicloud:
The formation was done via `find -type f -name  *.tf | xargs -L1 terraform fmt`


ketzacoatl:
Let's hold on to this until the next round of PRs are merged, we can then run a big format, and then work out updates to remaining PRs after/separately.

ketzacoatl:
We can open a new one.

PR #264, opened 2019-10-07
add a module to associate aws_iam_access_key credentials to a tfe wor…

JoseD92:
I add a module for associating a aws_iam_access_key credentials to a tfe workspace

### DIFF
--- a/modules/tf-cloud-credential/variables.tf
+++ b/modules/tf-cloud-credential/variables.tf
@@ -1,0 +1,22 @@
+variable "workspace_name_prefix" {
+  type        = string
+  description = "The name prefix to use for the workspace"
+}
+
+variable "organization" {
+  type        = string
+  description = "The workspace organization"
+}
+
+variable "iam_access_key" {
+  type = object({
+    id     = string
+    secret = string
+  })
+  description = "The aws_iam_access_key to use as credentials for workspace."
+}
+
+variable "region" {
+  type        = string
+  description = "The aws region"
+}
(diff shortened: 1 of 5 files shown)


modules/tf-cloud-credential/variables.tf
```diff
@@ -0,0 +1,22 @@
+variable "workspace_name_prefix" {
```
ketzacoatl: This would probably be better as `name_prefix`.

modules/tf-cloud-credential/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+  type        = string
+  description = "The workspace organization"
+}
+
+variable "iam_access_key" {
+  type = object({
+    id     = string
+    secret = string
+  })
+  description = "The aws_iam_access_key to use as credentials for workspace."
```
ketzacoatl: Should be: `"The aws_iam_access_key id/secret pair to use as credentials for the workspace."`

modules/tf-cloud-credential/variables.tf
mcgirr: I resolved this in the latest commit for this PR

modules/tf-cloud-credential/variables.tf
mcgirr: I also resolved this in the latest commit for this PR

mcgirr:
@JoseD92 I didn't want to hold up this PR but in the future when you make a PR for `terraform-aws-foundation` please be sure to update the release changelog with the changes the PR makes. Especially in this case since it adds a new module.

JoseD92:
Thanks @mcgirr, I will do next time.

PR #266, opened 2019-10-11
Credstash grant

### LINKED ISSUE #265: credstash-grant-reader could not be planned with empty(default) context_keys and context_values
Checking out the code, this line looks weird to me. https://github.com/fpco/terraform-aws-foundation/blob/d15768b3499638bba76c878e83659faa71741876/modules/credstash-grant-reader/main.tf#L54

Seems like I must pass one and only one key/value pair for each role. Wondering if this code was not as it supposed to be.

Magicloud:
Passed mock data planning. Fixes #265.

### DIFF
--- a/modules/credstash-grant-reader/main.tf
+++ b/modules/credstash-grant-reader/main.tf
@@ -24,16 +24,10 @@ variable "reader_policy_arn" {
   type        = string
 }
 
-variable "context_keys" {
-  default     = []
-  description = "list of keys to be zipped with the context_values to set an 'encryption context' for additional granularity that clients are required to provide to read encrypted values. Eg. for env=dev svc=db, this would be [env, svc]. All readers get this context map."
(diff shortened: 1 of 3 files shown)


ketzacoatl:
Should we also drop `role_count` and refactor/cleanup those variables all the way?

cc @Magicloud / @qrilka

qrilka:
@ketzacoatl yes, that was one of my proposals, just put it into a list of objects with names and ARNs. The current approach is quite awkward and error prone

Magicloud:
I am with removing "role_count". But seeing that comment, "circumvent issue", the issue is still open (for a few years....). @ketzacoatl Is that issue fixed or unrelated now, especially we are on 0.12 now.

ketzacoatl:
@Magicloud what are next steps for this PR?

PR #269, opened 2019-10-18
asg module: Default value for variable key_name

psibi:
The default value is an empty string which indicates the instance in
the ASG has no key associated with it. This makes the key to be optionally passed when using the module.

- [x] Update the changelog


ketzacoatl:
I believe we don't have a `default` value here b/c we want to force the module user to define that parameter, and in doing so, we also ensure they are explicit in the case when there is no key (which is not common unless using windows).

psibi:
Fair point!

PR #273, opened 2019-11-20
New modules: ALB / NLB / RDS

Magicloud:

- [x] Update the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,15 @@
 
 ### Modules
 
+* `alb`: Skeleton of ALB function.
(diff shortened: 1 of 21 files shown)


Magicloud:
@ketzacoatl Please checkout the new alb* modules and alb-test example. See if this is what you expected.

ketzacoatl:
@Magicloud can we break each of these new modules into separate PRs? It would make it easier to provide a good review on each use case when looking at them individually.

Magicloud:
Sure. Please checkout !288, !289, !290.

PR #275, opened 2019-11-25
WIP: Add an output for the role arn to the single-node-asg module.

mcgirr:
- This PR adds the `asg_iam_role_arn` output to the `single-node-asg` module
- This PR also makes a fix for the `single-node-asg` module by removing the `azs` variable which is no longer a part of the parent `asg` module since https://github.com/fpco/terraform-aws-foundation/pull/247

---

- [ ] Update the changelog

### DIFF
--- a/modules/single-node-asg/main.tf
+++ b/modules/single-node-asg/main.tf
@@ -91,4 +91,5 @@ module "init-attach-ebs" {
   source = "../init-snippet-attach-ebs-volume"
   region = var.region
   volume_id = module.service-data.volume_id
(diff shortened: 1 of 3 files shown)


ketzacoatl:
@mcgirr can you paste/screenshot/etc some output that confirm this works as expected? Is there an update to `tests/main.tf`?

mcgirr:
@ketzacoatl Sure thing. And I will remove the WIP at that point as well.

mcgirr:
This change is no longer needed.

PR #277, opened 2019-12-03
Fix iam-users output.

qrilka:
When a user gets deleted from user list Terraform will fail
creating a plan complaining that number of keys doesn't match
number of values.
Th solution is taken from https://github.com/terraform-google-modules/terraform-google-cloud-storage/pull/23/files

Example error seen:
```
  19:   value       = zipmap(var.user_list, aws_iam_user.u.*)
    |----------------
    | aws_iam_user.u is tuple with 8 elements
    | var.user_list is list of string with 7 elements

Call to function "zipmap" failed: number of keys (7) does not match number of
values (8).
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Modules
 
+* `iam-users`: fixed error from zipmap in outputs when a user gets deleted
+  from user list
 
 ### Examples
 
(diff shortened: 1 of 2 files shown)


ketzacoatl:
Confirmed in slack.

qrilka:
It looks like using `count` is a problem here and it makes sense to switch to using `for_each` instead with a set of user names

PR #278, opened 2019-12-04
Remove deprecated interpolation-only exprs from tf-cloud-credential

qrilka:
This is more or less trivial fix and it will remove a lot of deprecation warnings in our terraform runs

### DIFF
--- a/modules/tf-cloud-credential/main.tf
+++ b/modules/tf-cloud-credential/main.tf
@@ -1,5 +1,5 @@
 resource "tfe_variable" "workspace_aws_access_key_id" {
-  workspace_id = "${tfe_workspace.workspace.id}"
+  workspace_id = tfe_workspace.workspace.id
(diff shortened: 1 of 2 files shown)


qrilka:
@ketzacoatl how could I trigger CI?

ketzacoatl:
> This is more or less trivial fix and it will remove a lot of deprecation warnings in our terraform runs

Yea, LGTM.

> how could I trigger CI?

If the module has been added to `tests/main.tf`, pushing commits to a branch is all that should be needed.

ketzacoatl:
I believe I have added this module to the test suite, and the test suite is still green/passing (https://gitlab.fpcomplete.com/fpco-mirrors/terraform-aws-foundation/-/jobs/22949), but the script doesn't list the modules initialized/tested, so I'm unable to validate that I've added this module.

PR #281, opened 2019-12-11
module init-snippet-install-awscli: Uses Python3 for awscli.

Magicloud:
Python2 is deprecating.

---
name: Pull request template
---

- [ ] Update the changelog

### DIFF
--- a/modules/init-snippet-install-awscli/snippet.tpl
+++ b/modules/init-snippet-install-awscli/snippet.tpl
@@ -1,6 +1,5 @@
 # start snippet - install awscli python cli tools
 ${init_prefix}
-apt-get install -y python-pip
-pip install --upgrade pip==9.0.3
-pip install awscli
+apt install -y python3-pip
+pip3 install awscli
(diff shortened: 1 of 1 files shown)


modules/init-snippet-install-awscli/snippet.tpl
```diff
@@ -1,6 +1,6 @@
 # start snippet - install awscli python cli tools
 ${init_prefix}
-apt-get install -y python-pip
-pip install --upgrade pip==9.0.3
-pip install awscli
+apt update # in case this is missed in the whole user_data. Per my testing, APT won't download data updated within certain amount of time, except security related.
```
ketzacoatl: I'm not strongly objecting, but ideally, we would rely on the operator to do that, so this is not redundant. For example, if the operator is diligent, this module would punish them with extra apt updates than is needed.

Magicloud: OK.

PR #282, opened 2019-12-11
module init-snippet-attach-ebs-volume: Retrying attaching EBS

Magicloud:
In a rare case, the instance is running, but its AWS status is not "running". Then attaching process fails. The failure won't stop the instance hence ASG won't retry the whole process. This leaves us in an unexpected status.

This change is a patch up.

---
name: Pull request template
---

- [ ] Update the changelog


ketzacoatl:
@Magicloud in the case where the volume never attaches, this will block, loop and not complete, is that correct? I think that's ok, but we should make note of the limitation and scope out updates to address this in the future (I have some ideas).

PR #283, opened 2019-12-12
Support encrypted EBS in module ASG

Magicloud:
This is necessary for us to enable EBS encryption downstream.

### DIFF
--- a/modules/asg/variables.tf
+++ b/modules/asg/variables.tf
@@ -96,14 +96,14 @@ variable "root_volume_size" {
 variable "additional_block_devices" {
   default     = []
   description = "Additional EBS volumes to use as instance block devices"
-  type        = list(object({device_name = string, volume_type = string, volume_size = string }))
+  type        = list(object({ device_name = string, volume_type = string, volume_size = string, encrypted = bool }))
 }
 
 // List of maps, as extra tags to append to the Auto-Scaling Group
 variable "extra_tags" {
   default     = []
   description = "Extra tags that will be added to ASG, as a list of maps"
-  type        = list(object({key=string, value=string, propagate_at_launch=bool}))
+  type        = list(object({ key = string, value = string, propagate_at_launch = bool }))
   # see the example in this TF doc for more info:
   # https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html
   # should be provided in the following form:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Summary
 
 ### Modules
+* `asg`: Add EBS encryption support
 
 ### Examples
 
--- a/modules/asg/README.md
+++ b/modules/asg/README.md
@@ -16,7 +16,8 @@ The module supports:
   and `aws_role_arn` for setting up SNS
 * public IPs may be enabled/disabled
 * supports appending `extra_tags`
-* all important details (instance type, ami, key, user data, iam profile) are
+* all important details (instance type, ami, key, user data, iam profile, additional
+  volumes encryption) are
   specified as variables in the modules.
 
 Note that, Terraform does not run a rolling update when an ASG/LC pair have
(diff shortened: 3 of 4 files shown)


modules/asg/variables.tf
```diff
@@ -96,14 +96,14 @@ variable "root_volume_size" {
 variable "additional_block_devices" {
   default     = []
   description = "Additional EBS volumes to use as instance block devices"
-  type        = list(object({device_name = string, volume_type = string, volume_size = string }))
+  type        = list(object({ device_name = string, volume_type = string, volume_size = string, encrypted = bool }))
```
qrilka: There is a problem with optional attributes in HCL - https://github.com/hashicorp/terraform/issues/19898
If we specify it here then user will need to supply some value for `encrypted` (even if that value is `null`). As HCL is dynamically typed there is another way around: not specify optional attributes while still using them.

Magicloud: That is a good point. But, if I omit it here, hence user does not give a value for "encrypted". Then when I use it anyway, what would the value be?

qrilka: @Magicloud probably before I was doing something incorrectly but I don't have that code anymore. The current test:
```
variable "obj" {
  type = object({name = string, key = string})
}
resource "null_resource" "o" {
  provisioner "local-exec" {
    command = "echo ${var.obj.unspecified}"
  }
}
```
gives me an error:
```
Error: Unsupported attribute

  on main.tf line 26, in resource "null_resource" "o":
  26:     command = "echo ${var.obj.unspecified}"

This object does not have an attribute named "unspecified".
```
But replacing `var.obj.unspecified` with `lookup(var.obj, "unspecified", "default")` gives a better result:
```
$ terraform apply
var.obj
  Enter a value: {"name":"boo","key": 42}

null_resource.u[1]: Refreshing state... [id=2952313068434605786]
null_resource.o: Refreshing state... [id=5465750222682703737]
null_resource.u[0]: Refreshing state... [id=4296688519603364011]

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # null_resource.o is tainted, so must be replaced
-/+ resource "null_resource" "o" {
      ~ id = "5465750222682703737" -> (known after apply)
    }

Plan: 1 to add, 0 to change, 1 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

null_resource.o: Destroying... [id=5465750222682703737]
null_resource.o: Destruction complete after 0s
null_resource.o: Creating...
null_resource.o: Provisioning with 'local-exec'...
null_resource.o (local-exec): Executing: ["/bin/sh" "-c" "echo default"]
null_resource.o (local-exec): default
null_resource.o: Creation complete after 0s [id=1869318344897596251]
```

Magicloud: Great. Let me try some.

Magicloud: Not sure if I did something wrong. The default value did not work for me.
module: `encrypted   = lookup(ebs_block_device.value, "encrypted", false)`, caller: `additional_block_devices = [{device_name = "/dev/sdl", volume_type = "gp2", volume_size = 15, encrypted = true}]`. The generated plan and applied result was still not encrypted.

qrilka: yes, it looks like TF keeps only attributes you specify, that's very unfortunate :(
So then we could only specify everything explicitly (using `null` when we can't)

qrilka:
Plan with this version:

```
      - ebs_block_device { # forces replacement
          - delete_on_termination = true -> null
          - device_name           = "/dev/xvdb" -> null
          - encrypted             = false -> null
          - iops                  = 0 -> null
          - no_device             = false -> null
          - volume_size           = 250 -> null
          - volume_type           = "gp2" -> null
        }
      + ebs_block_device { # forces replacement
          + delete_on_termination = true
          + device_name           = "/dev/xvdb"
          + encrypted             = true
          + iops                  = (known after apply)
          + snapshot_id           = (known after apply)
          + volume_size           = 250
          + volume_type           = "gp2"
        }
```
And after applying it we have a new EBS volume shown as  encrypted

ketzacoatl:
Do we have any tests with the updated module using _non-encrypted_ EBS?

qrilka:
Updated README

PR #285, opened 2019-12-26
Asg root device encryption

qrilka:

- [x] Update the changelog

This follows #283 adding root encryption.

### DIFF
--- a/modules/asg/main.tf
+++ b/modules/asg/main.tf
@@ -92,6 +92,7 @@ resource "aws_launch_configuration" "cluster" {
   root_block_device {
     volume_type = var.root_volume_type
     volume_size = var.root_volume_size
+    encrypted   = var.root_encrypted
   }
 
   dynamic "ebs_block_device" {
@@ -108,4 +109,3 @@ resource "aws_launch_configuration" "cluster" {
     create_before_destroy = true
   }
 }
-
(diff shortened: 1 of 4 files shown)


qrilka:
Sample plan:
```
      ~ root_block_device {
            delete_on_termination = true
          ~ encrypted             = false -> true # forces replacement
          ~ iops                  = 0 -> (known after apply)
            volume_size           = 40
            volume_type           = "gp2"
        }
```
And AWS console shows root device as encrypted

ketzacoatl:
Seems like we should also update the `asg` module's `README.md` - https://github.com/fpco/terraform-aws-foundation/blob/asg-root-device-encryption/modules/asg/README.md

modules/asg/main.tf
```diff
@@ -92,6 +92,7 @@ resource "aws_launch_configuration" "cluster" {
   root_block_device {
     volume_type = var.root_volume_type
     volume_size = var.root_volume_size
+    encrypted = var.root_encrypted
```
ketzacoatl:
```suggestion
    encrypted   = var.root_encrypted
```

ketzacoatl: Whitespace alignment.

qrilka:
But we don't list all the available options - do we really need to say about root device encryption when we don't even say that we support root device size and type?

qrilka:
@ketzacoatl I've rebased this PR onto #283 and updated README so that one should be merged first

PR #288, opened 2020-01-20
New: module/rds

Magicloud:
Setup basic DB instance and subnet group resources for using RDS.

---
name: Pull request template
---

- [ ] Update the changelog

### DIFF
--- a/modules/rds/main.tf
+++ b/modules/rds/main.tf
@@ -1,0 +1,22 @@
+resource "aws_db_subnet_group" "rds_private_subnet" {
+  subnet_ids = var.subnet_ids
+}
+
+resource "aws_db_instance" "default" {
+  identifier_prefix       = var.name_prefix
+  allocated_storage       = var.db_storage_size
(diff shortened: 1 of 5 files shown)


modules/rds/main.tf
```diff
@@ -0,0 +1,105 @@
...
+  identifier_prefix       = var.name_prefix
+  allocated_storage       = var.db_storage_size
+  storage_type            = var.db_storage_type
+  engine                  = var.db_engine
+  engine_version          = var.engine_version
+  instance_class          = var.db_instance_type
+  db_subnet_group_name    = aws_db_subnet_group.rds_private_subnet.name
+  multi_az                = true
+  backup_retention_period = 7
+  monitoring_interval     = 30
```
ketzacoatl: These params here should be defined as a `variable` with a default value (or locals if there's a reason to not expose the knob, but I believe we want to expose those knobs).

Magicloud: Done.

PR #291, opened 2020-01-27
module/init-snippet-attach-ebs-volume: Manually make device name for attached EBS volume.

Magicloud:
Due to Linux kernel and Distro changes, an attched EBS volume may not get the specified device name (Ref https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html). Hence logic following may break since specified device does not exist.

This change uses Linux trick to manually setup (if it has not) the device name, so other logic can be static, without guessing which device is it.

---
name: Pull request template
---

- [ ] Update the changelog


ketzacoatl:
@Magicloud what are next steps for this PR?

PR #293, opened 2020-02-10
using workspace_id

JoseD92:
Changes tf-cloud-credential so that it receives the workspace_id as an argument instead of searching it with a datasource, because the datasource brings problems when the workspace does not exists

### DIFF
--- a/modules/tf-cloud-credential/data.tf
+++ b/modules/tf-cloud-credential/data.tf
@@ -1,4 +1,0 @@
-data "tfe_workspace" "workspace" {
-  name         = var.name_prefix
-  organization = var.organization
-}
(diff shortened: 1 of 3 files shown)


JoseD92:
The error for this PR happens when using this module and the code for making the tf workspace to which the module will add the variables to are within the same tf project. This is because the module uses a datasource to get the id of the workspace, but datasources only work on existing resources so both can't be add to the project simultaneously. Resulting in errors like:

```
Error: Could not find workspace My-Project/project_workspace

  on modules/tf-cloud-credential/data.tf line 1, in data "tfe_workspace" "workspace":
   1: data "tfe_workspace" "workspace" {
```

with the additions of this PR the plan is successful as it uses the id of the new workspace directly.

mcgirr:
Thanks for this PR @JoseD92 I will review it and get it merged

mcgirr:
This looks good to me. Thanks! Let's not remote the MR branch just yet until all the places that use it have been updated.

PR #301, opened 2020-02-28
A few changes to (mostly) single-node-asg module.

Magicloud:
While making confluence example, a few changes are needed:

Enable attach ELB to single node ASG. `single-node-asg` does not take ELB target group yet.

Remove dependency of Ubuntu on single-node-asg EBS attaching script. `ec2metadata` command exists only in Ubuntu AMI, and latest version for CentOS from AWS, not only has different name, but also output differently. So using raw http request to get rid of it.

Ensure EBS attachment ends up the device name we want.

Module/single-node-asg: Output attached data volume tag value "Name" so DLM can match the tag to manage the snapshots.

---
name: Pull request template
---

- [ ] Update the changelog

### DIFF
--- a/modules/single-node-asg/variables.tf
+++ b/modules/single-node-asg/variables.tf
@@ -115,9 +115,3 @@ variable "load_balancers" {
   description = "The list of load balancers names to pass to the ASG module"
   type        = list(string)
 }
-
-variable "alb_target_group_arns" {
-  default     = []
-  description = "list of target_group for application load balancer to associate with the ASG (by ARN)"
-  type        = list(string)
-}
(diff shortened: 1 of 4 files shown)


Magicloud:
@ketzacoatl Fixed.

modules/single-node-asg/variables.tf
```diff
@@ -116,3 +116,8 @@ variable "load_balancers" {
   type        = list(string)
 }
 
+variable "alb_target_group_arns" {
+  default     = []
+  description = "list of target_group for application load balancer to associate with the ASG (by ARN)"
+  type        = list(string)
+}
```
ketzacoatl: Let's review together when / where / how we associate ASGs with target groups, this can be tricky/problematic.

PR #302, opened 2020-02-29
Add kms to s3 remote state module

mcgirr:
Initial attempt to add server side encryption to the `s3-remote-state` module to support 

https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#server_side_encryption_configuration

for https://www.terraform.io/docs/backends/types/s3.html#kms_key_id

Contributes to https://github.com/fpco/terraform-aws-foundation/issues/310

- [ ] Update the changelog

### DIFF
--- a/modules/s3-remote-state/main.tf
+++ b/modules/s3-remote-state/main.tf
@@ -35,11 +35,25 @@ variable "force_destroy" {
   type        = bool
 }
 
+variable "kms_key_id" {
+  description = "The ARN of a KMS Key to use for encrypting the state"
+  type        = string
+}
+
 resource "aws_s3_bucket" "remote-state" {
   bucket        = var.bucket_name
   acl           = "private"
   region        = var.region
   force_destroy = var.force_destroy
+
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        kms_master_key_id = var.kms_key_id
+        sse_algorithm     = "aws:kms"
+      }
+    }
+  }
 
   versioning {
     enabled = var.versioning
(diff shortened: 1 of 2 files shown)


modules/s3-remote-state/main.tf
```diff
@@ -36,12 +35,26 @@ variable "force_destroy" {
   type        = bool
 }
 
+variable "kms_key_id" {
+  description = "The ARN of a KMS Key to use for encrypting the state"
+  type        = string
+}
+
```
ketzacoatl: The use of a KMS key here is a pattern worth encoding into a module like this. TY for this update.

I guess the complimentary update is a new module which defines the KMS key to be used by this module.

modules/s3-remote-state/main.tf
```diff
@@ -72,8 +85,6 @@ data "aws_iam_policy_document" "s3-full-access" {
   statement {
     effect = "Allow"
 
-    # find an authoritative list of valid Actions for a AWS bucket policy,
-    # I haven't been able to locate one, and the two commented out are invalid
```
ketzacoatl: Can we link to the authoritative list?

ketzacoatl: oh, I see it at the top, ignore this.

modules/s3-remote-state/main.tf
mcgirr: That sounds good to me. I've opened https://github.com/fpco/terraform-aws-foundation/issues/310 and included that detail in the issue.

modules/s3-remote-state/main.tf
mcgirr: :slightly_smiling_face:

PR #303, opened 2020-02-29
New module/init-snippet-install-docker-yum

Magicloud:
Redhat/Centos based initialization script to install `docker` and `docker-compose`.

---
name: Pull request template
---

- [ ] Update the changelog


modules/init-snippet-install-docker-yum/snippet.tpl
```diff
@@ -0,0 +1,6 @@
+yum install -y yum-utils device-mapper-persistent-data lvm2
+yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+yum install -y docker-ce docker-ce-cli containerd.io
+systemctl start docker
+curl -L "https://github.com/docker/compose/releases/download/1.25.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/bin/docker-compose
```
ketzacoatl: consider a variable for the version, which the module has a default value for.

PR #304, opened 2020-03-01
Enable attach ELB to single node ASG.

Magicloud:
This helps easily setup endpoint for ACM cert.

---
name: Pull request template
about: Make a PR to terraform-aws-foundation
---

Please include the following in your PR:

Please also note that these are not hard requirements, but merely serve to define
what maintainers are looking for in PR's.  Including these will more likely lead 
to your PR being reviewed and accepted.

- [ ] Update the changelog
- [ ] Make sure that modules and files are documented. This can be done inside the module and files.
- [ ] Make sure that new modules directories contain a basic README.md file.
- [ ] Make sure that the module is added to [tests/main.tf](https://github.com/fpco/terraform-aws-foundation/blob/master/tests/main.tf)
- [ ] Make sure that the linting passes on CI.
- [ ] Make sure that there is an up to date example for your code:
      - For new `modules` this would entail example code for how to use the module or some explanation in the module readme.
      - For new examples please provide a README explaining how to run the example. It's also ideal to provide a basic makefile to use the example as well.
- [ ] Make sure that there is a manual CI trigger that can test the deployment.

PR #307, opened 2020-03-04
Module/single-port-sg: Support source_security_group

Magicloud:
---
name: Pull request template
---

- [ ] Update the changelog


Magicloud:
> Is there a way to use this without a targeted apply? I'm running into this situation that I think is stemming from the changes in the PR:
> 
> ```
> count                    = local.udp * local.by_src_sg
> 
> The "count" value depends on resource attributes that cannot be determined
> until apply, so Terraform cannot predict how many instances will be created.
> To work around this, use the -target argument to first apply only the
> resources that the count depends on.
> ```

I feel like this is a dep issue. I got this while destroying resources again. The source SG was gone (and this rule), but TF says it cannot calculate the count.

The reason I do it this way is that, `source_security_group` and `cidr_blocks` cannot be set at the same time, even one is an empty value.

modules/single-port-sg/main.tf
```diff
@@ -40,30 +47,56 @@ variable "udp" {
...
   to_port           = var.port
   protocol          = "udp"
-  cidr_blocks       = var.cidr_blocks
   security_group_id = var.security_group_id
+  cidr_blocks       = var.cidr_blocks
+}
+
+# ingress rule for tcp, if enabled
+resource "aws_security_group_rule" "tcp_ingress_src_sg" {
+  count                    = local.tcp * local.by_src_sg
```
ketzacoatl: Define the logic/math as some local value we refer to here. For example, this is `enable_tcp_by_src_sg` (or `enable_tcp_by_sg`) and then there is `enable_tcp_by_cidr`

ketzacoatl:
@Magicloud please update the above, TY!

modules/single-port-sg/main.tf
Magicloud: Hi, I had a discussion in Gitlab. TL;DR, the whole function does not work well right now.

If the source security group does not exist yet, TF fails to calculate the numbers. Targeting the source SG first makes this work, but not ideal. So we need new design, for example, a different module to support source security group (and not calculating).

Magicloud:
Updated new design. Tested in PR https://github.com/fpco/terraform-aws-foundation/pull/304.

PR #319, opened 2020-04-03
Update Changelog for new release

Magicloud:
---
name: Pull request template
---

- [ ] Update the changelog


Magicloud:
@ketzacoatl The change works, but not the final code (filling up description and so) since I'd like to hear your idea. Is it better to make another few modules instead of making breaking changes to original modules?

modules/init-snippet-attach-ebs-volume/instance_id.tpl
```diff
@@ -0,0 +1,7 @@
+if which wget; then
+  INSTANCE_ID="$(wget -O- http://169.254.169.254/latest/meta-data/instance-id)"
+elif which curl; then
+  INSTANCE_ID="$(curl http://169.254.169.254/latest/meta-data/instance-id)"
+fi
+
```
ketzacoatl, line 6: If no curl and no wget, is this a silent failure with redherring elsewhere, or do we fail more loudly?

modules/init-snippet-attach-ebs-volume/snippet.tpl
```diff
@@ -1,13 +1,5 @@
...
 export AWS_DEFAULT_REGION=${region}
 VOLUME_ID=${volume_id}
-if which wget; then
-  INSTANCE_ID="$(wget -O- http://169.254.169.254/latest/meta-data/instance-id)"
-elif which curl; then
-  INSTANCE_ID="$(curl http://169.254.169.254/latest/meta-data/instance-id)"
-fi
-
 if [ "x$${INSTANCE_ID}" == "x" ]; then
   echo 'OS not functioning'
```
ketzacoatl, line 4: What would be "not functioning" here? Let's be more specific with the error message so an operator gets a more clear suggestion of what to do.

modules/init-snippet-attach-ebs-volume/snippet.tpl
```diff
@@ -30,5 +22,3 @@ else
   dev_name="/dev/$(readlink "$${dev_id}" | tr / '\n' | tail -1)"
   [ "$${dev_name}" == "${device_path}" ] || ln -s "$${dev_name}" "${device_path}"
 fi
-
-${init_suffix}
```
ketzacoatl, line 34: Why is this dropped?

modules/persistent-ebs/locals.tf
```diff
@@ -0,0 +1,11 @@
+locals {
+  volume_default = {
+    type        = "gp2"
+    iops        = 0
+    size        = 15
+    encrypted   = true
+    kms_key_id  = ""
+    snapshot_id = ""
+  }
+  volumes_default = [for x in var.volumes : merge(local.volume_default, x)]
```
ketzacoatl, line 10: Does this mean the module _always_ uses the default? I don't see how that's optional.

modules/persistent-ebs/main.tf
```diff
@@ -9,25 +9,27 @@
...
+  type              = local.volumes_default[count.index].type
+  encrypted         = local.volumes_default[count.index].encrypted
+  kms_key_id        = local.volumes_default[count.index].kms_key_id
+  snapshot_id       = local.volumes_default[count.index].snapshot_id
 
   # merge Name w/ extra_tags
   tags = merge(
     {
-      "Name" = "${var.name_prefix}-${var.az}"
+      "Name" = "${var.name_prefix}-${var.az}-${local.volumes_default[count.index].name}"
```
ketzacoatl, line 24: This is great, that LGTM.

modules/persistent-ebs/outputs.tf
```diff
@@ -1,5 +1,5 @@
-output "volume_id" {
-  value       = aws_ebs_volume.main.id
+output "volume_ids" {
+  value       = aws_ebs_volume.main.*.id
   description = "`id` exported from the `aws_ebs_volume`"
```
ketzacoatl, line 3: Should add `type = list(string)`.

Edit: oh, this is an output, ignore.

modules/persistent-ebs/variables.tf
```diff
@@ -60,3 +24,15 @@ variable "iam_instance_profile_role_name" {
...
+  type = list(object({
+    type        = string,
+    iops        = number,
+    size        = number,
+    encrypted   = bool,
+    kms_key_id  = string,
+    snapshot_id = string,
+    name        = string
+  }))
+}
```
ketzacoatl, line 38: Should define `description`. No default is good, this is a required parameter.

modules/single-node-asg/main.tf
```diff
@@ -21,12 +21,20 @@ locals {
...
   az = data.aws_subnet.server-subnet.availability_zone
 
-  # The "name_prefix" we provide to the persistent-ebs module
-  data_volume_name_prefix = "${var.name_prefix}-${var.name_suffix}-data"
-
   # The `name_prefix` we provide to the `iam-instance-profile` module
   # The persistent-ebs module appends the AZ to the data node name, other modules don't do that
   name_prefix_with_az = "${var.name_prefix}-${var.name_suffix}-${local.az}"
+
+  data_volume_default = {
```
ketzacoatl, line 28: why the `_default` in the name?

modules/single-node-asg/outputs.tf
```diff
@@ -7,8 +7,3 @@ output "asg_iam_role_name" {
   value       = module.instance_profile.iam_role_name
   description = "`name` exported from the Service Data `aws_iam_role`"
 }
-
-output "data_volume_name_tag" {
-  value       = "${local.data_volume_name_prefix}-${local.az}"
-  description = "Name tag value for attached data volume"
-}
```
ketzacoatl, line 14: Should this be kept, or why drop it?

modules/single-node-asg/variables.tf
```diff
@@ -115,3 +79,15 @@ variable "load_balancers" {
...
+  type = list(map(any))
+  #    type = string,
+  #    iops = number,
+  #    size = number,
+  #    encrypted = bool,
+  #    kms_key_id = string,
+  #    snapshot_id = string,
+  #    name = string,
+  #    device = string
+}
```
ketzacoatl, line 93: Did the map spec fail, or why is it commented out here?

modules/init-snippet-attach-ebs-volume/instance_id.tpl
Magicloud: This is handled in snippet.tpl. But I will move it here.

modules/init-snippet-attach-ebs-volume/snippet.tpl
Magicloud: OK.

modules/init-snippet-attach-ebs-volume/snippet.tpl
Magicloud: This part (and init_prefix) is moved to the output part. So this snippet is focused on attaching EBS.

modules/persistent-ebs/locals.tf
Magicloud: No, this merges user value and default. Without this, user must specify all values of the map, of each element (EBS volume) of the list. Now he just needs to specify the necessary ones.

modules/single-node-asg/main.tf
Magicloud: It means the default value of the map. In variable, we cannot specify default values for list(map(any)). So this is the way to reach the same goal (along with the merge call below).

modules/single-node-asg/outputs.tf
Magicloud: I am thinking a better way to do this. The value is for DLM to match the volume. Since we are having multiple now. The output does not work.

modules/single-node-asg/variables.tf
Magicloud: I designed this as object, but it did not work as expected with `merge`, so converted to map.

ketzacoatl:
@Magicloud do we also want to create a copy of the previous `single-node-asg` module to allow users access to the previous workflow?

qrilka:
@Magicloud what's the status of this? Do you plan to add README as requested above and add a module to use with the previous workflow?

Magicloud:
> Please add a README and Makefile with the example, and include in the README notes on how to use the example and run the tests.

Done.

modules/single-node-asg/main.tf
```diff
@@ -82,8 +85,10 @@ module "server" {
   user_data = <<END_INIT
 #!/bin/bash
 # exec > /tmp/init.log
-# exec 2> /tmp/init-err.log
+exec 2>&1
 # set -x
+apt update
+${module.install-awscli.init_snippet}
```
qrilka, line 91: Please update header comments as well (about CLI availability), thanks

qrilka:
I tried to create an ASG with no extra volumes (passing `[]`) got an error during apply:
```

Error: Invalid index

  on .terraform/modules/asg-a/modules/init-snippet-attach-ebs-volume/main.tf line 19, in data "template_file" "init_snippet":
  19:     device_path   = var.device_paths[count.index]
    |----------------
    | count.index is 0
    | var.device_paths is empty list of string

The given key does not identify an element in this collection value.
```

Magicloud:
> I tried to create an ASG with no extra volumes (passing `[]`) got an error during apply:
> 
> ```
> 
> Error: Invalid index
> 
>   on .terraform/modules/asg-a/modules/init-snippet-attach-ebs-volume/main.tf line 19, in data "template_file" "init_snippet":
>   19:     device_path   = var.device_paths[count.index]
>     |----------------
>     | count.index is 0
>     | var.device_paths is empty list of string
> 
> The given key does not identify an element in this collection value.
> ```

@qrilka Does this happen on existing resources? I tried from beginning, no problem.

qrilka:
@Magicloud yes, probably that was the reason - I switched from `master` version to this branch

qrilka:
So probably it  could be a problem when upgrading modules (and when this will get merged)

Magicloud:
@qrilka I see. Since the logic is changed totally, it could break.

qrilka:
@Magicloud while using this branch for A/B switching demo I have found that tags cause unnecessary updates like the following:
```
  # module.asg-magenta.module.service-data.aws_ebs_volume.main[0] will be updated in-place
  ~ resource "aws_ebs_volume" "main" {
        arn                  = "arn:aws:ec2:us-west-2:xxxxx:volume/vol-xxxxx"
        availability_zone    = "us-west-2a"
        encrypted            = true
        id                   = "vol-xxxx"
        iops                 = 100
        kms_key_id           = "arn:aws:kms:us-west-2:xxxx:key/xxxx-ae63-4278-bba4-xxxx"
        multi_attach_enabled = false
        size                 = 20
      ~ tags                 = {
            "Name"             = "switching-demo-magenta-us-west-2a-a"
          - "cldy-instance-id" = "i-0b4f3320df090f7ca" -> null
        }
        type                 = "gp2"
    }
```

Magicloud:
@qrilka Where does that tag come from?

qrilka:
That's a good question but I have not idea and not sure if I will have time to find it out...

ketzacoatl:
That tag is unrelated and we should not have blocked this PR on that. Unless there's reason to hold back, I think we should merge this PR.

PR #320, opened 2020-04-07
Switch CI to Github Actions

snoyberg:


ketzacoatl:
cc @mcgirr / @Magicloud

mcgirr:
> It sounds like github actions require using a docker image, and we will not have a default build image with our tools, so we should include the docker image in the repo as well?

I'm getting up to speed with Github actions and will have a better response soon but if this is the case then yes I think that makes sense.

snoyberg:
This PR includes a workflow for building the Docker image used by the lint workflow.

mcgirr:
I'm more familiar with GitHub Workflows/Actions now. I could shift back to review this PR if that's wanted.

ketzacoatl:
@mcgirr, sure, it would be good to bring CI back for this repo.

PR #322, opened 2020-04-14
Extend nat-gateways modules to enable support for EIP switching

psibi:
This is a backward compatible change with no updates needed to
existing modules. The additional variables can be used for scenarions
where migrating a EIP from one NAT to another is required.

- [x] Update the changelog

### DIFF
--- a/modules/nat-gateways/main.tf
+++ b/modules/nat-gateways/main.tf
@@ -9,10 +9,21 @@
  *
  */
 
+locals {
+  total_nat_count = var.enable_nat_creation ? var.nat_count : 0
+  total_new_nat   = var.enable_nat_creation ? (length(var.nat_eip) == 0 ? local.total_nat_count : 0) : 0
+  nat_ids         = var.enable_nat_creation ? (length(var.nat_eip) == 0 ? aws_eip.nat.*.id : values(data.aws_eip.nat)[*].id) : []
+}
+
 # AWS Managed NAT Gateways
 resource "aws_eip" "nat" {
-  count = var.nat_count
+  count = local.total_new_nat
   vpc   = true
+}
+
+data "aws_eip" "nat" {
+  for_each  = length(var.nat_eip) != 0 ? toset(var.nat_eip) : toset([])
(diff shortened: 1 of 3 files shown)


modules/nat-gateways/variables.tf
```diff
@@ -29,3 +29,14 @@ variable "extra_tags" {
...
+  default     = true
+  description = "boolean, enable/disable NAT creation"
+  type        = string
+}
+
+variable "nat_eip" {
+  description = "The public IP of the specific EIP to retrieve. If non empty, this list should have same number of EIP as the number of var.public_subnet_ids."
+  type        = list(string)
+  default     = []
+}
```
ketzacoatl, line 42: If empty list, AWS will create EIP and associate with the managed NAT Gateways on their own. If the list includes EIP, we lookup and tell AWS to use those for the NAT Gateways to be created. Is that correct?

modules/nat-gateways/main.tf
```diff
@@ -9,21 +9,32 @@
  *
  */
 
+locals {
+  total_nat_count = var.enable_nat_creation ? var.nat_count : 0
+  total_new_nat   = var.enable_nat_creation ? (length(var.nat_eip) == 0 ? local.total_nat_count : 0) : 0
+  nat_ids         = var.enable_nat_creation ? (length(var.nat_eip) == 0 ? aws_eip.nat.*.id : values(data.aws_eip.nat)[*].id) : []
+}
```
ketzacoatl, line 16: please add a comment above each local value to help future maintainers understand the intention/purpose of each of those.

modules/nat-gateways/variables.tf
psibi: Yes, that's right.

modules/nat-gateways/main.tf
psibi: @ketzacoatl I have added comments to each of the variable and also renamed `nat_ids` to `eip_ids` for better clarity. Thanks for the review!

PR #325, opened 2020-04-28
Ec2 connect tunnel

qrilka:

- [x] Update the changelog

This PR uses #232 and #319 so needs those to be merged first


modules/ec2-connect-tunnel/outputs.tf
```diff
@@ -2,3 +2,27 @@ output "public_ip" {
...
+  description = "`name` exported from the Service Data `aws_iam_role`"
+}
+
+output "data_volume_name_tag" {
+  value       = module.asg.data_volume_name_tag
+  description = "Name tag value for attached data volume."
+}
+
+output "eip_address" {
+  value = module.asg.eip_address
```
qrilka, line 27: @JoseD92 this output already exists as `public_ip`

